PR #75, opened 2023-10-06
74 update iojsonwebtokenjjwt

pwannenmacher:
close https://github.com/Frachtwerk/essencium-backend/issues/74


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Version ___
```
cr0ssing, line 3: Version number is missing

MIGRATION.md
```diff
@@ -1,5 +1,9 @@
 # Migration Guide
 
+## Migrate to `___`
```
cr0ssing, line 3: Version number is missing

cr0ssing (changes requested):

CHANGELOG.md
pwannenmacher: The version number is still missing here on purpose. It is not yet certain whether the next release will be `2.4.6` or `2.5.0`. Before I push the corresponding tag the migration guide and changelog will be adjusted accordingly. The entry here is more a reminder...

MIGRATION.md
pwannenmacher: The version number is still missing here on purpose. It is not yet certain whether the next release will be `2.4.6` or `2.5.0`. Before I push the corresponding tag the migration guide and changelog will be adjusted accordingly. The entry here is more a reminder...

pwannenmacher:
As long as we don't have a solution to persist a signing key over several Essencium instances in parallel (e.g. in load balancer scenarios), the secret must be set via an ENV variable. Therefore the latest changes.

cr0ssing (approved):

PR #146, opened 2023-11-16
feat: switch to multiple roles per user

pwannenmacher:


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/initialization/DefaultUserInitializer.java
```diff
@@ -19,85 +19,74 @@
...
+                    .ifPresentOrElse(
+                        user -> {
+                          // clear roles
+                          user.setRoles(new HashSet<>());
+                          userService.save(user);
+
+                          // update user with new roles
+                          user.setFirstName(userProperties.getFirstName());
+                          user.setLastName(userProperties.getLastName());
+                          user.setPassword(userProperties.getPassword());
```
pwannenmacher, line 72: Should we really overwrite a chosen password?

pwannenmacher: should only be set during initialisation

mrAtFW (approved):

PR #328, opened 2024-03-11
107 remove default success controller

hoelger-frachtwerk:


hoelger-frachtwerk:
In the response(-detail) you can still see that the last instance checking for some existing resource is from `StaticResourceConfig.java` ... not sure if this should be somewhat more generic

```
{
    "type": "about:blank",
    "title": "Not Found",
    "status": 404,
    "detail": "No static resource thisshouldfail.",
    "instance": "/thisshouldfail"
}
```

pwannenmacher (commented):
The file [`essencium-backend/src/main/resources/public/index.html`](https://github.com/Frachtwerk/essencium-backend/blob/b08819ae9519c390ab6d47e4753d0aaf3da860cd/essencium-backend/src/main/resources/public/index.html) is therefore superfluous and should also be removed.

hoelger-frachtwerk:
Screenshot_20240319_113220
You still can access the file and I think it's usefull as a basic test "is the server running". Although we have the health endpoint, this one is more obvious.

pwannenmacher (approved):

PR #341, opened 2024-03-15
99 enforce basic password policy for new users

hoelger-frachtwerk:


pwannenmacher (commented):
There are a total of 3 changes in this PR:
- Changing the default password strength
- Extension of the JSON parameters of the `PasswordUpdateRequest`
   - `newPassword` as alias for `password
   - `resetToken` and `oldPassword` as alias for `verification`
- Non-essential change to the language variables for error messages regarding passwords that are too weak
As the names of the JSON parameters of the password change endpoint are extended, there are no breaking changes for API consumers.
Perhaps we should add the information about the levels and the default setting in the documentation `/docs`.

hoelger-frachtwerk:
Now added documentation for `min-password-strength`

pwannenmacher (approved):

PR #497, opened 2024-06-30
feat(deps): make allowSignup, updateRole, userRoleAttr and roles conf…

pwannenmacher:
…igurable per provider

### DIFF
--- a/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/oauth2/OAuth2SuccessHandler.java
+++ b/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/oauth2/OAuth2SuccessHandler.java
@@ -41,6 +41,7 @@ import java.util.*;
 import lombok.RequiredArgsConstructor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.security.authentication.ProviderNotFoundException;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.security.oauth2.client.authentication.OAuth2AuthenticationToken;
@@ -95,6 +96,10 @@ public class OAuth2SuccessHandler<
 
       final OAuth2ClientRegistrationProperties.ClientProvider clientProvider =
           oAuth2ClientRegistrationProperties.getProvider().get(providerName);
+      if (Objects.isNull(clientProvider)) {
+        throw new ProviderNotFoundException(
+            "could not resolve client provider for provider '" + providerName + "', aborting");
+      }
 
       UserInfoEssentials userInfo;
       try {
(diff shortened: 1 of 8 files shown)


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/oauth2/OAuth2SuccessHandler.java
```diff
@@ -90,67 +90,75 @@ public void onAuthenticationSuccess(
...
+        redirectHandler.onAuthenticationSuccess(request, response, authentication);
+        return;
       }
 
-      userService.patch(Objects.requireNonNull(user.getId()), patch);
-
-      redirectHandler.setToken(tokenService.createToken(user, SessionTokenType.ACCESS, null, null));
-    } catch (UsernameNotFoundException e) {
-      LOGGER.info("user {} not found locally", userInfo.getUsername());
+      try {
```
JelmenGuhlke, line 112: The `try-and-catch` block should only be around the relevant code lines. For here only the `userService.loadUserByUsername(userInfo.getUsername());` can throw the `UsernameNotFoundException`. For a better readability, only this line should be covered by the try-and-catch block. The two possible flows (found vs. not found) can also be capsulated in two separate methods.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/oauth2/OAuth2SuccessHandler.java
```diff
@@ -90,67 +90,75 @@ public void onAuthenticationSuccess(
...
-        Role defaultRole = roleService.getDefaultRole();
-        if (roles.isEmpty() && Objects.nonNull(defaultRole)) {
-          LOGGER.info("no roles found for user '{}'. Using default Role.", userInfo.getUsername());
-          roles.add(defaultRole);
-        }
-        patch.put("roles", roles);
+    if (authentication instanceof OAuth2AuthenticationToken oAuth2AuthenticationToken) {
+      final String providerName = oAuth2AuthenticationToken.getAuthorizedClientRegistrationId();
+
+      final OAuth2ClientRegistrationProperties.ClientProvider clientProvider =
```
JelmenGuhlke, line 96: Can the result of `oAuth2ClientRegistrationProperties.getProvider().get(providerName);` be `null`? From my point of view, if the result is`null` an Exception should be thrown right after the call

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/oauth2/OAuth2SuccessHandler.java
```diff
@@ -179,6 +187,8 @@ private UserInfoEssentials extractUserInfo(
       OAuth2AuthenticationToken authentication, String providerName)
       throws UserEssentialsException {
     final UserInfoEssentials userInfo = new UserInfoEssentials();
+    OAuth2ClientRegistrationProperties.ClientProvider clientProvider =
```
JelmenGuhlke, line 190: Can the result of oAuth2ClientRegistrationProperties.getProvider().get(providerName); be null? From my point of view, if the result isnull an Exception should be thrown right after the call

JelmenGuhlke (changes requested):

Issue #502, opened 2024-07-03
Use `jwtAuthenticationProvider` only in `JwtTokenAuthenticationFilter`


JelmenGuhlke:
**Is your feature request related to a problem? Please describe.**
During some profiling action I investigate a heavy CPU time consumption in the `DaoAuthenticationProvider#additionalAuthenticationChecks` method of Spring. The problem here is, that the `DaoAuthenticationProvider` is part of the `JwtTokenAuthenticationFilter` authentication managers. So even if a JWT is present in the HTTP header, the `DaoAuthenticationProvider` hashes the given input (the JWT as `String`) and compares it with the users password.

**Describe the solution you'd like**
During the `JwtTokenAuthenticationFilter` only the `jwtAuthenticationProvider` should be applied. If there is no match, all the other providers can be run.

PR #503, opened 2024-07-03
feat(auth): use correct JWT provider in filter

JelmenGuhlke:
- #502
- only use the `jwtAuthenticationProvider` in `JwtTokenAuthenticationFilter` to prevent unnecessary hashing and comparisons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Version `_________`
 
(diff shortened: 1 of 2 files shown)


pwannenmacher (approved):
Tested using external and internal Authentication Providers. `DaoAuthenticationProvider` is only called once on each Login. When authenticating via JWT only the `JwtAuthenticationProvider` is called (as desired).

PR #579, opened 2024-09-04
feat(BaseUserSpec): email filter accepts partial emails

cholakov11:
Users can be filtered by using only a substring of an email.

* Defined new Specification PartialEmailSpecBase
* Wrote integration test to check filtering by full and partial emails

### DIFF
--- a/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/repository/specification/BaseUserSpec.java
+++ b/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/repository/specification/BaseUserSpec.java
@@ -46,5 +46,12 @@ interface NameSpecBase<USER extends AbstractBaseUser<ID>, ID extends Serializabl
 interface EmailSpecBase<USER extends AbstractBaseUser<ID>, ID extends Serializable>
     extends BaseModelSpec<USER, ID> {}
 
+@Spec(
+    path = "email",
+    params = {"email", "user", "username"},
+    spec = LikeIgnoreCase.class)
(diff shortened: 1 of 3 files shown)


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/repository/specification/BaseUserSpec.java
```diff
@@ -46,5 +46,12 @@ interface NameSpecBase<USER extends AbstractBaseUser<ID>, ID extends Serializabl
 interface EmailSpecBase<USER extends AbstractBaseUser<ID>, ID extends Serializable>
```
JelmenGuhlke, line 46: please remove the old, unsed spec interface

essencium-backend/src/testIntegration/java/de/frachtwerk/essencium/backend/test/integration/controller/UserControllerIntegrationTest.java
```diff
@@ -725,6 +725,69 @@ void testFilterUsersByName() throws Exception {
         .andExpect(jsonPath("$.content.[0].id", is(user2.getId()), Long.class));
   }
 
+  @Test
+  void testFilterUsersByEmail() throws Exception {
```
JelmenGuhlke, line 729: For a better readability: you can nested your tests with `@Nested` and a human readable name with `@DisplayName`.

Furthermore, we should only test one case per test method

JelmenGuhlke (changes requested):
Please add a short change hint in the `CHANGELOG.md` file

JelmenGuhlke (approved):
:+1:

PR #580, opened 2024-09-04
refactor: AbstractUserController extends AbstractAccessAwareController

### LINKED ISSUE #388: AbstractUserController does not inherit from AbstractAccessAwareController
Some functionalities have to be implemented twice, because the AbstractUserController does not inherit from AbstractAccessAwareController. When implementing access to "owned" entities, this requires specifications in the method signatures. 

AbstractUserController (and the RoleController) should inherit from AbstractAccessAwareController. 

The outcome might be to not implement this, then it should be documented why. There might be some useful discussions about this in the old Gitlab repo.

**additional context**
"Nicht ID als Long sondern als Teil der Specification" is part of my notes but I would have to get more into the code, to understand what paul wanted to say with this.

cholakov11:
AbstractUserController now extends the AbstractAccessAwareController

* Removed redundant methods
* Implemented abstract methods from AbstractAccessAwareController
* Adjusted methods to use inherited functionality

closes #388

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,8 +1,20 @@
 # Migration Guide
 
+## Version `______`
+
+- The `updateObject` method in the `AbstractUserController` class has been renamed to `update`. Every call to the `updateObject(ID, USERDTO, SPEC)` method of the `AbstractUserController` should be replaced with `update(ID, USERDTO, SPEC)`.
+
+Example:
+```java
+userController.updateObject(id, userDTO, spec);
+```
+to
+```java
+userController.update(id, userDTO, spec);
+```
 ## Version `2.7.0`
 
-- All implementations of the `AbstractUserController` should be checked to see whether one of the methods `findById`, `updateObject`, `update`, `delete` or, `terminate` have been overwritten. If this is the case, the method signature must be adapted to the use of specifications. 
+- All implementations of the `AbstractUserController` should be checked to see whether one of the methods `findById`, `updateObject`, `update`, `delete` or, `terminate` have been overwritten. If this is the case, the method signature must be adapted to the use of specifications.
 
 Example:
 ```java
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -2,21 +2,33 @@
 
 ## Version `______`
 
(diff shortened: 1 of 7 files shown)


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractAccessAwareController.java
```diff
@@ -209,6 +209,7 @@ public List<BasicRepresentation> findAll(@Parameter(hidden = true) SPEC specific
       required = true,
       content = @Content(schema = @Schema(type = "integer")))
   public REPRESENTATION findById(
+      @PathVariable("id") @NotNull final ID id,
```
JelmenGuhlke, line 212: Maybe we were talking past each other: we don't need the ID as a parameter here, so it can be removed. We're willing to accept the "unbind"-hint from the IDE/Sonar.

MIGRATION.md
```diff
@@ -1,8 +1,20 @@
...
+userController.updateObject(id, userDTO, spec);
+```
+to
+```java
+userController.update(id, userDTO, spec);
+```
 ## Version `2.7.0`
 
-- All implementations of the `AbstractUserController` should be checked to see whether one of the methods `findById`, `updateObject`, `update`, `delete` or, `terminate` have been overwritten. If this is the case, the method signature must be adapted to the use of specifications. 
+- All implementations of the `AbstractUserController` should be checked to see whether one of the methods `findById`, `updateObject`, `update`, `delete` or, `terminate` have been overwritten. If this is the case, the method signature must be adapted to the use of specifications.
```
JelmenGuhlke, line 17: Which method signature of the `AbstractUserController` has changed?

JelmenGuhlke (changes requested):

MIGRATION.md
cholakov11: This is not written by me. It is actually from the previous Essencium version 2.7.  I am also not sure what  is meant here.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractAccessAwareController.java
cholakov11: Id parameter is removed again.

JelmenGuhlke (dismissed):
LGTM

PR #589, opened 2024-09-16
fix: 390-handling-property-reference-exception

### LINKED ISSUE #390: Wrong values in SORT BY lead to error status 500
**Describe the bug**
From a child project I have the error `org.springframework.data.mapping.PropertyReferenceException: No property '<string>' found for type 'Zone' in "findAllZones"` because SORT BY expects the arguments column and order (`cathegory|partner|hub, desc|asc`). 

**To Reproduce**
TODO

**Expected behavior**
When unexpected values are received, this should be a "bad request" with error status 400 and a more verbose error message, so the user directly knows what to improve.

**Additional context**
A solution is probably to be implemented in `de.frachtwerk.essencium.backend.controller.adviceRestExceptionHandler` and with 

code snippet from my notes:
```
@ExceptionHandler(PropertyReferenceException.class)
protected ResponseEntity<Object> handlePropertyReferenceException(
  PropertyReferenceException ex, WebRequest request) {
    return new ResponseEntity<>();
}
```

cholakov11:
Implemented an exception handler to handle PropertyReferenceException caused by an invalid value in the `sort` query parameter. Previous behaviour resulted in an `Internal Error 500` response with a full trace of the exception. Now the API will respond with the error format used for reporting other errors (e.g Unauthorized) and omits the exception trace thus resulting in a more clear and user-friendly response.

*Note:* Invalid values for the other query parameters - page and size, are handled with an OK response from the server and a json containing results for the request with the invalid values substituted by default valid. In order to keep the invalid query handling consistent, the initial idea was to filter the `sort` parameter for invalid values and in case of such present to return an unsorted response.
This could be achieved with reflection and matching `sort` parameter with defined fields of the java object. However since the Pageable interface is used to transmit the query parameters, the Pageable object annot be edited in a pre-processing function to remove invalid sort values and pass a new Pageable with only valid values and thus avoid the PropertyReferenceException. The only option is to construct a `SimplePageRequest` which implements `Pageable` with the edited sort values and pass it on. This however breaks the Dependency Inversion principle and should be avoided.
Therefore, at this point no solution will be implemented to handles the sort parameter like the other two and we will resolve to an ExceptionHandler.

fixes #390

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,13 @@
 
 ## Version `______`
 
-- Feature: `BasicRepresentation` can be used to create slim representations of entities. `BasicRepresentation` is very well suited for filling embedded objects or lists of objects in other representations.
-- Feature: Each Controller now has a `GET .../basic` endpoint that returns a List of BasicRepresentations of the entity. Filtering via Specifications is possible
+- Feature: `BasicRepresentation` can be used to create slim representations of entities. `BasicRepresentation` is very
+  well suited for filling embedded objects or lists of objects in other representations.
+- Feature: Each Controller now has a `GET .../basic` endpoint that returns a List of BasicRepresentations of the entity.
+  Filtering via Specifications is possible
 - Feature: `AbstractUserController` can now filter users by using a substring of an email.
+- fix: Invalid values for the sort query, resulting in PropertyReferenceException are handled by an ExceptionHandler and
+  result in a "Bad Request" response.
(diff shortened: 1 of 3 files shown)


JelmenGuhlke:
@cholakov11 can we close this PR in favor of https://github.com/Frachtwerk/essencium-backend/pull/606 ?

pwannenmacher (changes requested):
CHANGELOG.md conflicts with main. Some unwanted changes seem to have occurred there (probably due to automatic formatting). Please rebase and remove all unwanted changes to CHANGELOG.md.

PR #606, opened 2024-09-30
593-exception-handling

### LINKED ISSUE #593: Framework related Exception handling
In general it is a good practise to implement domain / application specific Exceptions based on default JVM Exceptions like `IllegalArgumentException`  or `InvalidStateException`. This approach has several advantages:

- Precise error identification: Custom exception classes provide more specific and meaningful error messages.
- Better maintainability: They make debugging and future maintenance easier by clearly indicating where and why issues occur.
- Clean abstraction: Custom exceptions encapsulate framework-specific details, making the framework easier to use without exposing internal workings.
- Improved error handling: Users of the framework can handle different types of exceptions with more granularity and flexibility.
- Self-documenting code: Custom exceptions make the codebase clearer and help with understanding error contexts at a glance.
- Enhanced API documentation: Clearly defined exceptions improve the framework's documentation, making it easier for users to know what to expect.

Essencium has currently in the package `de.frachtwerk.essencium.backend.model.exception` only a few framework related implementations for `Exception` classes. These are partly used. However, JVM standard exceptions are still being thrown in some cases.

To make exception handling easier for application projects with Essencium, it should be ensured that all thrown exceptions are custom implementations of exception classes. These should be enriched with use-case relate context information, e.g. for `InvalidInputException` at least the invalid raw input should be present.

**To-Do**
- [ ] implement meaningfull `Exception` classes, with included context related data
- [ ] make sure, that only custom `Exception` classes are manually thrown
- [ ] add `Exception` class model with a short descriptions to the `README.md` file
- [ ] investigate options for handling (or conscious not handling) Spring framework related exceptions, like validation `Exeption` cases or Spring Data exception

cholakov11:
Closes #593


pwannenmacher:
The license block is missing from all the newly introduced classes.

pwannenmacher:
The merge conflicts with main need to be resolved.

pwannenmacher (changes requested):
The license block is missing from all the newly introduced classes.

pwannenmacher (changes requested):
The merge conflicts with main need to be resolved.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/exception/ResourceActions.java
```diff
@@ -0,0 +1,19 @@
+package de.frachtwerk.essencium.backend.model.exception;
+
+public enum ResourceActions {
```
pwannenmacher, line 3: If we are using an enum as a source for static strings, do we also want to use a final set of strings? For example, a ResourceAction “IMPORT” or “CHECK” is conceivable. I would either leave out this hard coding or use the HTTP methods as a set.

pwannenmacher (changes requested):

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractAccessAwareController.java
```diff
@@ -209,8 +209,10 @@ public List<BasicRepresentation> findAll(@Parameter(hidden = true) SPEC specific
       required = true,
       content = @Content(schema = @Schema(type = "integer")))
   public REPRESENTATION findById(
-      @Parameter(hidden = true) @Spec(path = "id", pathVars = "id", spec = Equal.class) SPEC spec) {
-    return toRepresentation(service.getOne(spec).orElseThrow(ResourceNotFoundException::new));
+      @Parameter(hidden = true) @Spec(path = "id", pathVars = "id", spec = Equal.class) SPEC spec,
+      @PathVariable ID id) {
+    return toRepresentation(
+        service.getOne(spec).orElseThrow(() -> new ResourceCannotFindException(id.toString())));
```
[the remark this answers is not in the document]
cholakov11: I haven't actually changed any program flow in this PR. Only the exceptions strucutre is changed and old exceptions were replaced with the new ones. Anyway, I believe that in this case the exception throwing should be moved to the service and then stick to this convetion for future changes.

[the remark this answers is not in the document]
cholakov11: > Wouldn't something similar have to happen with PUT, PATCH and DELETE?

So the other methods have first the `testAccess` call which performs one test and then throws a specification exception. Then on the next step the respective `update/create/patch` method is called which has its pre- and post- functions where I believe we should leave for the developer to specify how the program flow is handled (exception throwing, tolerating, etc.). When we move the exception throwing out of the findById call then we won't have any more exception handling logic in the controller and don't need to update PUT, PATCH and DELETE methods.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/exception/ResourceActions.java
cholakov11: This enum is actually used purely as an util in order to avoid writing manually everywhere the different actions. The exceptions structure itself is not limited to this enum. In the end the action which caused an exception is stored as a String in the respective exception and not as an instance of ResourceActions. I am fine with removing it since there won't be a big difference. We can  input directly as strings the actions where required.

cholakov11:
> The license block is missing from all the newly introduced classes.

License is now added.

> The merge conflicts with main need to be resolved.

Merge conflicts resolved

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/exception/ResourceActions.java
cholakov11: What is the current status of this discussion? Do we decide for removing the enum or keeping it?

cholakov11:
I don't know why but my permission to access and push to the repository is being denied. I cannot push the last commits which handle maintainability issues and increase the test coverage.

PR #607, opened 2024-10-01
fix: removed redundant @NotNull annotations

### LINKED ISSUE #388: AbstractUserController does not inherit from AbstractAccessAwareController

cholakov11:
Closes #388


JelmenGuhlke (approved):
:+1: LGTM

PR #677, opened 2025-02-27
fix: added @NotNull constraints to AbstractUserController parameters

### LINKED ISSUE #673: AbstractUserController redefines parameter constraints of AbstractAccessAwareController
**Describe the bug**
A class that inherits from the AbstractUserController and overrides the create, update or delete method always violates the parameter constraint configuration because the AbstractAccessAwareController uses `@NotNull` annotations

**To Reproduce**
1. Create a class that inherits from AbstractUserController and overrides one of its methods
2. Send a request on the endpoint of the new class

**Expected behavior**
Don't return a 500 "HV000151: A method overriding another method must not redefine the parameter constraint configuration, but method Class#method(Parameter) redefines the configuration of AbstractAccessAwareController#method(Parameter)."

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser Firefox
 - Version 135.0.1

**Additional context**
Occured when updating jourfixe to Essencium 2.10.0

trusty-fire:
closes #673

### DIFF
--- a/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractUserController.java
+++ b/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractUserController.java
@@ -274,7 +274,7 @@ public abstract class AbstractUserController<
   @Secured({BasicApplicationRight.Authority.USER_CREATE})
   @ResponseStatus(HttpStatus.CREATED)
   @Operation(summary = "Create a new user")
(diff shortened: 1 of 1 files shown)


mrAtFW (approved):
Looks fine, but are they all not implicitly set by the path variable?

trusty-fire:
> Looks fine, but are they all not implicitly set by the path variable?

There are errors when you try to override the methods because the constraints do not match the super method

PR #699, opened 2025-04-28
604 basicrepresentation does not allow null values as name

### LINKED ISSUE #604: BasicRepresentation does not allow `null` values as `name`
**Describe the bug**
The current implementations of the `BasicRepresentation#from` factory methods returns silently `null`, if the given `name` parameter is `null`:

```java
  public static BasicRepresentation from(Serializable id, String name) {
    if (Objects.isNull(id) || Objects.isNull(name)) {
      return null;
    }
    return new BasicRepresentation(id, name);
  }
```

**To Reproduce**
1. Call `BasicRepresentation.from(1, null)` from any class of the application
2. The returned `value` is `null`

**Expected behavior**
A `BasicRepresentation` is used as a DTO object during a REST request. From the client perspective an object / entity with an `id` is *existing* - regardless if the `name` is `null`. For this reason, a `BasicRepresentation` should be valid, if only the `id` is present.

viridispirus:


pwannenmacher:
@viridispirus Just a quick note: Could you add this behavior change to the CHANGELOG?

PR #721, opened 2025-06-17
feat: implement sorting of users by name parameter

lwydra:
This PR adjusts the `AbstractUserService` to replace potential sorting parameters with `name` by `firstName` and `lastName`, aligning the sorting- and filter-parameters for users.


pwannenmacher (commented):
Both the request `GET {{baseUrl}}/v1/users?sort=firstName,lastName` (old and new implementation) and the request `{{baseUrl}}/v1/users?sort=name` (new implementation) lead to the following SQL query:

```sql
select u1_0."id",
       u1_0."created_at",
       u1_0."created_by",
       u1_0."email",
       u1_0."enabled",
       u1_0."failed_login_attempts",
       u1_0."first_name",
       u1_0."last_name",
       u1_0."locale",
       u1_0."login_disabled",
       u1_0."mobile",
       u1_0."nonce",
       u1_0."password",
       u1_0."password_reset_token",
       u1_0."phone",
       u1_0."source",
       u1_0."updated_at",
       u1_0."updated_by"
from "FW_USER" u1_0
order by u1_0."first_name", u1_0."last_name"
offset ? rows fetch first ? rows only;
```

Why is the separate query parameter needed again?

lwydra:
After offline discussion with @pwannenmacher it has been decided to move the implementation of the sorting parameter into the demo-specific code, as more of the user-attributes may be moved from essencium-core into the demo-application in the future

PR #818, opened 2025-11-23
817 upgrade to spring boot 40x

### LINKED ISSUE #817: Upgrade to Spring Boot 4.0.x

pwannenmacher:


essencium-backend/src/testIntegration/java/de/frachtwerk/essencium/backend/test/integration/util/AbstractEssenciumIntegrationTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+import org.testcontainers.postgresql.PostgreSQLContainer;
+import org.testcontainers.utility.DockerImageName;
+
+@Testcontainers
+@ActiveProfiles({"test_postgresql"})
+public abstract class AbstractEssenciumIntegrationTest {
+  public static final DockerImageName POSTGRES_IMAGE = DockerImageName.parse("postgres:18");
+
+  static final PostgreSQLContainer POSTGRES_CONTAINER = new PostgreSQLContainer(POSTGRES_IMAGE);
```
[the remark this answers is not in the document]
pwannenmacher: false positive

essencium-backend-development/src/main/resources/application-h2.yaml
```diff
@@ -21,6 +21,8 @@ spring:
     driver-class-name: org.h2.Driver
     username: sa
     password: password
-
   jpa:
     database-platform: org.hibernate.dialect.H2Dialect
+    properties:
+      hibernate:
+        globally_quoted_identifiers: true
```
jnsfZ: ## H2 profiles silently disable Hibernate 7 snake_case conversion.

- **Hibernate config drift across H2/test profiles.** `hibernate.globally_quoted_identifiers: true` is set in three files: `essencium-backend-development/src/main/resources/application-debug_h2.yaml`, `essencium-backend-development/src/test/resources/application-test.yaml`, and `essencium-backend/src/test/resources/application-local_test.yaml`. The main dev profile `application-h2.yaml` correctly uses `auto_quote_keyword: true`.

  **Effect**: under those three profiles, Hibernate 7's `PhysicalNamingStrategySnakeCaseImpl.quotedIdentifier()` returns names verbatim — so entities like `ApiToken`, `SessionToken`, `ExampleEntity` produce CamelCase tables (`FW_APITOKEN`, etc.) instead of snake_case (`FW_API_TOKEN`). The dev workflow is unaffected (the main `h2` profile is correct), but `mvn test` runs against a different schema layout than `mvn spring-boot:run` would. Inconsistency that's easy to overlook.

  **Why the setting exists at all**: H2 rejects unquoted `key` / `value` columns (`Translation.key`, `Translation.value`, `SessionToken.key`); PG accepts them as non-reserved keywords. The Hibernate 7 docs recommend the targeted alternative — *"`globally_quoted_identifiers` … is not recommended"*, use `auto_quote_keyword` instead.

  - **Fix**: in each of the three files, replace

    ```yaml
    spring.jpa.properties.hibernate.globally_quoted_identifiers: true
    ```

    with

    ```yaml
    spring.jpa.properties.hibernate.auto_quote_keyword: true
    ```

    matching what `application-h2.yaml` already does. Reserved-word columns stay quoted; everything else flows through the unquoted path and gets snake-cased.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/access/AccessAwareJsonFilter.java
```diff
@@ -19,50 +19,74 @@
...
   }
 
+  @Override
+  public void depositSchemaProperty(
+      PropertyWriter writer, JsonObjectFormatVisitor objectVisitor, SerializationContext context) {
+    // Default implementation - include all properties in schema
+  }
+
+  private boolean isAllowed(Object value, PropertyWriter writer) {
+    JsonAllowFor ann = writer.getMember().getAnnotation(JsonAllowFor.class);
```
jnsfZ, line 73: Nice to have: 
- **No integration test for the `@JsonAllowFor` filter pipeline.** The pipeline (`AccessAwareJsonViewAdvice.determineWriteHints` → `JacksonJsonHttpMessageConverter` → `AccessAwareJsonFilter`) is the role-based JSON filter. Its correctness for admin vs. non-admin responses, list responses, and missing-principal cases is uncovered by any test.
  - **Fix**: add an IT with a DTO annotated `@JsonFilter("roleBasedFilter")` and a field `@JsonAllowFor(roles = {"ADMIN"})`. Cases: (a) admin sees field, (b) non-admin response omits it, (c) `List<Dto>` to lock in per-element behavior.

essencium-backend/src/testIntegration/resources/db/migration/V1_0__init.sql
```diff
@@ -1,4 +1,4 @@
-create table api_token
+create table if not exists api_token
```
jnsfZ, line 1: Got this finding, maybe worth adressing in a follow up ticket:
- **`essencium-backend/src/testIntegration/resources/db/migration/V1_0__init.sql`** -- missing `native` / `foreign` tables. `AccessEntityFilteringIntegrationTest` only passes because `ddl-auto: update` runs after Flyway and backfills them. A flip to `validate` breaks the build with no log signal.
  - **Fix**: add `native`, `foreign`, and the join tables to `V1_0__init.sql`. Or document the `ddl-auto: update` reliance in the migration.

essencium-backend/src/testIntegration/java/de/frachtwerk/essencium/backend/test/integration/model/NativeIdIntegrationTest.java
```diff
@@ -1,80 +0,0 @@
-/*
```
jnsfZ, line 1: Nice to have:
- **Lost coverage from deleted `NativeIdIntegrationTest`** -- it asserted that `IDENTITY`-strategy primary keys aren't silently shared across two entity types (a real Hibernate quirk). `AbstractBaseModelIntegrationTest` only covers audit fields.
  - **Fix**: port the assertion into `AbstractBaseModelIntegrationTest` or a new test:

    ```java
    @Test
    void identityIdsAreNotSharedAcrossEntityTypes() {
      Native n = nativeRepository.save(new Native(...));
      Foreign f = foreignRepository.save(new Foreign(...));
      assertThat(n.getId()).isNotEqualTo(f.getId());
    }
    ```

essencium-backend/pom.xml
```diff
@@ -114,39 +116,41 @@
...
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-freemarker</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-beans</artifactId>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-jackson</artifactId>
```
jnsfZ: - **`essencium-backend/pom.xml`: `spring-boot-starter-jackson` is redundant** -- already pulled transitively by `spring-boot-starter-webmvc`. Remove:

  ```xml
  <dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-jackson</artifactId>
  </dependency>
  ```

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtAuthenticationProvider.java
```diff
@@ -36,6 +34,8 @@
 import org.springframework.security.authentication.dao.AbstractUserDetailsAuthenticationProvider;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.userdetails.UserDetails;
+import tools.jackson.core.type.TypeReference;
+import tools.jackson.databind.ObjectMapper;
 
 /** Provider to fetch user details for a previously extracted and validated JWT token */
```
jnsfZ, line 40: Nur sure if the bug is introduced in this pr, tested with development:
- **`POST /auth/renew` returns 500 NullPointerException** — the refresh-token flow is broken. Calling `POST /auth/renew` with the `refreshToken` cookie + `Authorization: Bearer <access>` triggers:

  ```
  java.lang.NullPointerException: Cannot invoke "java.util.Collection.stream()"
    because the return value of "io.jsonwebtoken.Claims.get(Object)" is null
    at JwtAuthenticationToken.createPrincipal(JwtAuthenticationToken.java:65)
  ```

  The refresh JWT only carries `sub` / `iat` / `exp` / `iss` claims — no `roles` or `rights`. `createPrincipal` reads them unconditionally via `Claims.get("roles").stream()` and NPEs on a null result. Effect: access tokens cannot be renewed; clients get logged out at the 15-minute access-token expiry with no path forward except a full re-login.

  - **Fix**: in `JwtAuthenticationToken.createPrincipal`, guard the claim reads — either treat missing `roles` / `rights` as empty collections (refresh tokens shouldn't grant authorities), or branch on the token `typ` claim (`REFRESH` vs `ACCESS`) and skip authority population entirely for refresh tokens. The renew handler should be looking up the user's *current* authorities from the database, not reading them from the refresh-token payload.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/AccessAwareJsonViewAdvice.java
```diff
@@ -19,37 +19,70 @@
...
-      @NotNull ServerHttpResponse response) {
-    EssenciumUserUtil.getUserDetailsFromAuthentication()
-        .ifPresent(
-            principal -> {
-              FilterProvider filters =
-                  new SimpleFilterProvider()
-                      .addFilter(FILTER_NAME, new AccessAwareJsonFilter<>(principal));
-              bodyContainer.setFilters(filters);
-            });
+  public boolean supports(
```
jnsfZ, line 44: `supports()` substring-matches: `converterName.contains("Jackson") && converterName.contains("HttpMessageConverter")`. Matches every Jackson dataformat converter on the classpath (XML, CBOR, Smile, YAML) — they ignore the JSON-only `FilterProvider` hint, so output is correct, but the advice still builds a `SimpleFilterProvider` per request for converters that won't use it. Also brittle to Spring class renames.
  - **Fix**:

    ```java
    @Override
    public boolean supports(MethodParameter returnType,
                            Class<? extends HttpMessageConverter<?>> converterType) {
      return JacksonJsonHttpMessageConverter.class.isAssignableFrom(converterType);
    }
    ```

essencium-backend/pom.xml
```diff
@@ -245,7 +249,7 @@
         <!-- https://mvnrepository.com/artifact/io.sentry/sentry-spring-boot-starter-jakarta -->
         <dependency>
             <groupId>io.sentry</groupId>
-            <artifactId>sentry-spring-boot-starter-jakarta</artifactId>
+            <artifactId>sentry-spring-boot-4</artifactId>
```
jnsfZ, line 253: - **`sentry-spring-7` dependency missing** -- `essencium-backend/pom.xml` declares `sentry-spring-boot-4` but not the standalone `sentry-spring-7` artifact. The Sentry/Spring integration is split across two dependencies in the new Sentry packaging; without `sentry-spring-7`, Spring-MVC-specific instrumentation (`@SentryTransaction`, MVC handler tracing) is unavailable.
  - **Fix**: add the matching `io.sentry:sentry-spring-7` dependency at the same version as `sentry-spring-boot-4`.

jnsfZ (changes requested):

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/dto/EssenciumUserDetails.java
```diff
@@ -53,7 +54,7 @@ public class EssenciumUserDetails<ID extends Serializable>
   public static final Locale DEFAULT_LOCALE = Locale.GERMAN;
 
   @Override
-  public Collection<? extends GrantedAuthority> getAuthorities() {
+  public @Nonnull Collection<? extends GrantedAuthority> getAuthorities() {
```
jnsfZ, line 57: Could be also adjusted in a follow up Ticket:
- **Join-table names and FK columns retain CamelCase under PG / H2.** Live DDL from a clean dev-app startup:

  ```
  create table "FW_APITOKEN_RIGHTS" (api_token_id uuid not null, "rights_authority" varchar(255) not null, …)
  create table "FW_USER_ROLES"      ("User_id"  bigint not null, roles_name varchar(255) not null, …)
  ```

  Both should be `FW_API_TOKEN_RIGHTS` / `user_id`. The main entity tables (`FW_API_TOKEN`, `FW_USER`) are correctly snake-cased, but the `@JoinTable` / `@JoinColumn` defaults derived from them inherit the already-quoted parent identifier and hit Hibernate 7's `quotedIdentifier()` short-circuit — same root cause as the H2-profile drift above, but in a different code path.

  - **Fix**: add explicit `@JoinTable(name = "fw_api_token_rights", joinColumns = @JoinColumn(name = "api_token_id"))` and `@JoinTable(name = "fw_user_roles", joinColumns = @JoinColumn(name = "user_id"))` to the `@ManyToMany` declarations in `ApiToken` and `AbstractBaseUser`. Or fix the upstream cause by overriding `toPhysicalTableName` / `toPhysicalColumnName` in `DataNamingConfig` to also process quoted identifiers through the snake_case path.

jnsfZ (changes requested):

essencium-backend-development/src/main/resources/application-h2.yaml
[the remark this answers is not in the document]
pwannenmacher: addressed in https://github.com/Frachtwerk/essencium-backend/pull/818/commits/d5ab5b5a96fb175b20ec9799fab522b3f0bb3b8a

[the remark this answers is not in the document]
pwannenmacher: addressed in https://github.com/Frachtwerk/essencium-backend/pull/818/commits/d5ab5b5a96fb175b20ec9799fab522b3f0bb3b8a

pwannenmacher: addressed in https://github.com/Frachtwerk/essencium-backend/pull/818/commits/d5ab5b5a96fb175b20ec9799fab522b3f0bb3b8a

essencium-backend/src/testIntegration/resources/db/migration/V1_0__init.sql
pwannenmacher: fixed in https://github.com/Frachtwerk/essencium-backend/pull/818/commits/56330c7fbb8ccbb2229438e23ef10e53c624903a

essencium-backend/pom.xml
pwannenmacher: fixed in https://github.com/Frachtwerk/essencium-backend/pull/818/commits/e1d400be6618ad617e50c6eb8562a6d818d983a3

essencium-backend/pom.xml
pwannenmacher: `sentry-spring-boot-4` bindet sehr wohl `sentry-spring-7` ein.  `io.sentry:sentry-spring-boot-4` hängt transitiv von `io.sentry:sentry-spring-7:8.43.2 ab (Scope: compile)

```xml
  <dependencies>
    <dependency>
      <groupId>io.sentry</groupId>
      <artifactId>sentry</artifactId>
      <version>8.43.2</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>io.sentry</groupId>
      <artifactId>sentry-spring-7</artifactId>
      <version>8.43.2</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>io.sentry</groupId>
      <artifactId>sentry-reactor</artifactId>
      <version>8.43.2</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
```

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/dto/EssenciumUserDetails.java
pwannenmacher: Wasn't that already fixed with https://github.com/Frachtwerk/essencium-backend/pull/818#discussion_r3386727140?

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/AccessAwareJsonViewAdvice.java
pwannenmacher: fixed by https://github.com/Frachtwerk/essencium-backend/pull/818/commits/873bfb992386004e35533ea981809d0f5e4c3d85

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/access/AccessAwareJsonFilter.java
pwannenmacher: Addressed by `AccessEntityFilteringIntegrationTest.testJsonAllowForShowsAdminOnlyPropertyForAdmin` and `AccessEntityFilteringIntegrationTest.testJsonAllowForHidesAdminOnlyPropertyForNonAdmin`. added additional UnitTest `AccessAwareJsonViewAdviceTest`

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/dto/EssenciumUserDetails.java
jnsfZ: Tested it, no its not. somehow hibernate works different for the manytomany tables. It still produces this without the JoinTable: 

  Under globally_quoted_identifiers: true:
  ```
    "FW_APITOKEN_RIGHTS" ("ApiToken_id" ...)
    "FW_USER_ROLES"      ("User_id" ...)
    "FW_SESSIONTOKEN"    (... "parentToken_id" ...)
 ```
 ```

  Under auto_quote_keyword: true:
    "FW_APITOKEN_RIGHTS" (api_token_id ...)              <- table still CamelCase, FK column fixed
    "FW_USER_ROLES"      ("User_id" ...)                  <- FK column still CamelCase
    "FW_SESSION_TOKEN"   (... parent_token_id ...)        <- both fixed
  ```


Both could be fixed by adding explicit @JoinTable(name="fw_api_token_rights",  JoinColumns=@JoinColumn(name="api_token_id")) on ApiToken.rights and  @JoinTable(name="fw_user_roles", joinColumns=@JoinColumn(name="user_id")) on AbstractBaseUser.roles.

jnsfZ (dismissed):

jnsfZ (approved):

PR #848, opened 2025-12-01
chore: add common interface for AbstractBaseUser and EssenciumUserDet…

lwydra:
This PR adds a common interface to `AbstractBaseUser` and `EssenciumUserDetails`.
This will be useful when an application needs to perform permission checks both on the current user via `EssenciumUserDetails`, but also on arbitrary `User` entities from the database, as now the checks can operate on the common interface instead of having to be duplicated for both classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Version `3.3.0-SNAPSHOT`
+
+### 🌟 Features
+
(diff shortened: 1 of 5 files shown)


pwannenmacher (commented):
Would you like to store the method `getAdditionalClaimByKey(String key, Class<O> clazz)` in `BaseEssenciumUserDetails` as part of this PR?

pwannenmacher (approved):

PR #860, opened 2026-04-14
feat: add configuration switches to enable/disable mail delivery for various mail types

pwannenmacher:
…

### DIFF
--- a/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/ContactMailService.java
+++ b/essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/ContactMailService.java
@@ -53,13 +53,19 @@ public class ContactMailService<
 
   public void sendContactRequest(
       @NotNull final ContactRequestDto contactRequest, final AUTHUSER issuingUser) {
-    try {
-      final Mail mailToSend =
-          buildMailFromRequest(sanitizeUserInformation(contactRequest, issuingUser));
-      mailService.sendMail(mailToSend);
-    } catch (IOException | TemplateException e) {
-      Sentry.captureException(e);
-      log.error("Error while sending contact request mail", e);
+    if (contactMailConfig.isEnabled()) {
+      try {
+        final Mail mailToSend =
+            buildMailFromRequest(sanitizeUserInformation(contactRequest, issuingUser));
+        mailService.sendMail(mailToSend);
+      } catch (IOException | TemplateException e) {
+        Sentry.captureException(e);
+        log.error("Error while sending contact request mail", e);
+      }
+    } else {
+      log.warn(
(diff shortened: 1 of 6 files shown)


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/ContactMailService.java
```diff
@@ -53,13 +54,20 @@ public class ContactMailService<
...
+        mailService.sendMail(mailToSend);
+      } catch (IOException | TemplateException e) {
+        Sentry.captureException(e);
+        log.error("Error while sending contact request mail", e);
+      }
+    } else {
+      log.warn(
+          "Contact mail service is disabled. Contact request from {} has not been sent.",
+          Optional.ofNullable(contactRequest.getMailAddress())
+              .orElse(Optional.ofNullable(issuingUser.getUsername()).orElse("N/A")));
```
jnsfZ: Can `issuingUser` be `null` here? If yes, this throws a NPE. `.orElse()` always evaluates its argument up front, so `issuingUser.getUsername()` gets called even when `getMailAddress()` is present. `orElseGet(() -> ...)` would fix this by only evaluating the fallback when needed.

```suggestion
log.warn(
    "Contact mail service is disabled. Contact request from {} has not been sent.",
    Optional.ofNullable(contactRequest.getMailAddress())
        .orElseGet(() -> Optional.ofNullable(issuingUser)
            .map(EssenciumUserDetails::getUsername)
            .orElse("N/A")));
```

essencium-backend/src/test/java/de/frachtwerk/essencium/backend/service/ContactMailServiceTest.java
```diff
@@ -22,10 +22,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.reset;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
jnsfZ: was the wildcard import intentional?

jnsfZ (dismissed):

essencium-backend/src/test/java/de/frachtwerk/essencium/backend/service/ContactMailServiceTest.java
pwannenmacher: nope 😅

jnsfZ (approved):

Issue #870, opened 2026-04-15
fix: ErrorResponse @Value injection silently does nothing


jnsfZ:
## Problem

The `ErrorResponse` class uses `@Value("${server.error.include-message}")` to control whether error messages should be included in API error responses. However, `ErrorResponse` is not a Spring bean -- it is created with `new ErrorResponse(...)` in the `RestExceptionHandler`. Spring's `@Value` injection only works for objects managed by the Spring container.

The field `includeMessageMode` is therefore **always null**, and the `includeMessage()` method always returns `true`. The configuration `server.error.include-message` is completely ignored.

## Context

- **File:** [`ErrorResponse.java`](essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/dto/ErrorResponse.java), lines 41-56

```java
@Value("${server.error.include-message}") // not a Spring bean -- injection never fires
private ErrorProperties.IncludeAttribute includeMessageMode;

private boolean includeMessage() {
    return includeMessageMode == null // always true since field is always null
        || includeMessageMode.equals(ErrorProperties.IncludeAttribute.ALWAYS);
}
```

- **Caller:** [`RestExceptionHandler.java`](essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandler.java) (all places where `new ErrorResponse(...)` is called)
- **Impact:** Downstream projects that set `server.error.include-message=NEVER` (to avoid leaking internal details in production) are **not actually protected** -- error messages are always exposed.

## Implementation Hint

Inject the `ErrorProperties` into `RestExceptionHandler` and pass the `includeMessage` setting to the `ErrorResponse` constructor instead of relying on `@Value` on a non-bean.

## Acceptance Criteria

- [ ] The configuration `server.error.include-message` is respected
- [ ] When set to `NEVER`, error responses do not contain detailed error messages
- [ ] When set to `ALWAYS` (or default), behavior is unchanged
- [ ] `CHANGELOG.md` updated
- [ ] `MIGRATION.md` updated (if the fix changes behavior for downstream projects)

deepika1214:
Hi @jnsfZ, I’d like to take this issue on. Could you please assign it to me? Let me know if there’s anything I should review before getting started.

jnsfZ:
> Hi [@jnsfZ](https://github.com/jnsfZ), I’d like to take this issue on. Could you please assign it to me? Let me know if there’s anything I should review before getting started.

[...]

PR #877, opened 2026-04-16
#870 fix: respect server.error.include-message configuration in ErrorResponse

deepika1214:
### Solution
- Inject `ServerProperties` into `RestExceptionHandler`, which  is a Spring-managed bean
- Evaluate `includeMessage` once at construction time
- Pass the resolved `boolean` into the `ErrorResponse` constructor
- Remove the dead `@Value` field and `includeMessage()` method from  `ErrorResponse` entirely

### Changes
| File | Change |
|------|--------|
| `ErrorResponse.java` | Removed `@Value` field and `includeMessage()` method; constructor now accepts `boolean includeMessage` |
| `RestExceptionHandler.java` | Added `ServerProperties` constructor injection; passes `includeMessage` to all 5 `ErrorResponse` call sites |


jnsfZ:
Could u also please add some Kind of Integration Test to verify that the new Code works?

pwannenmacher:
replaced by https://github.com/Frachtwerk/essencium-backend/pull/912

PR #879, opened 2026-04-19
fix(users): enforce ownership access control on terminate endpoint

### LINKED ISSUE #862: fix: terminate endpoint bypasses ownership-based access control
## Problem

The `POST /v1/users/{id}/terminate` endpoint (forces a user logout by invalidating all their session tokens) does not check ownership-based access control. Any user with the `USER_UPDATE` permission can terminate sessions of **any other user** -- even if ownership restrictions are configured that should limit users to managing only their own resources.

All other mutating endpoints in the same controller (update, delete, etc.) use the `spec` parameter for access control. The `terminate` endpoint receives this parameter but never uses it.

## Context

- **File:** [`AbstractUserController.java`](essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractUserController.java), lines 347-364

```java
// spec is injected but never used -- no access control check
public void terminate(@PathVariable @NotNull final ID id, ... SPEC spec) {
    userService.terminate(super.service.getById(id).getUsername());
}
```
- **Impact:** Affects downstream projects using ownership-based access control (`@RestrictAccessToOwnedEntities`). Projects relying only on `@Secured` are not affected since the endpoint already requires `USER_UPDATE`.
- **Comparison:** The `delete` endpoint directly above (lines 341-344) shows the correct pattern

## Acceptance Criteria

- [ ] The `terminate` endpoint enforces access control like all other mutating endpoints
- [ ] A user with ownership restrictions can only terminate their own sessions
- [ ] Test covering this behavior
- [ ] `CHANGELOG.md` updated
- [ ] `MIGRATION.md` updated (if the fix changes behavior for downstream projects)

tmchow:
## Summary

Closes #862. `POST /v1/users/{id}/terminate` received the `Spec` parameter for ownership-based access control but never used it, so any user with `USER_UPDATE` could terminate sessions of any other user even when `@RestrictAccessToOwnedEntities` was configured. Every other mutating endpoint in `AbstractUserController` (update, patch, delete, deleteToken, ...) routes through `service.testAccess(spec)` first. This PR brings `terminate` in line with that pattern.

## Change

`essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/AbstractUserController.java`:

```diff
   public void terminate(
       @PathVariable @NotNull final ID id,
       @Spec(path = "id", pathVars = "id", spec = Equal.class) @Parameter(hidden = true) SPEC spec) {
-    userService.terminate(super.service.getById(id).getUsername());
+    userService.terminate(super.service.testAccess(spec).getById(id).getUsername());
   }
```

The `delete` endpoint directly above at line 343 uses the same shape -- `service.testAccess(spec).deleteById(id)`. When `testAccess` rejects (ownership filter excludes the target user) it throws `ResourceNotFoundException`, which is the same response surface the other mutating endpoints already produce for the same scenario.

## Tests

Both controller test classes (`UUIDUserControllerTest`, `LongUserControllerTest`):

1. **Existing `terminate()` test** -- updated to stub `testAccess(spec)` and verify it's invoked. Without the stub the new chain call would hit a `NullPointerException` on the mock.

2. **New `terminateDeniedByOwnershipSpecThrows()` test** -- stubs `testAccess` to throw `ResourceNotFoundException`, asserts the exception propagates, and asserts `getById` / `terminate` are never reached. This is the regression pin that blocks the bug from coming back.

The `testAccess(spec).thenReturn(userServiceMock)` pattern is already used by the existing `update` / `patch` tests in both files (see UUIDUserControllerTest.java line 149), so the new setup mirrors established style.

## Docs

- **`CHANGELOG.md`** -- 3.4.0-SNAPSHOT "🐞 Bug Fixes" section gets a bullet describing the fix.
- **`MIGRATION.md`** -- new `Version 3.4.0` section explaining the downstream impact:
  - Projects using `@Secured` only (`USER_UPDATE`): unaffected.
  - Projects using `@RestrictAccessToOwnedEntities`: `terminate` will now respect the ownership restriction; clients that were previously allowed to terminate another user's session will now receive `ResourceNotFoundException` (same response `delete` / `update` would have returned under the same conditions).
  - No schema change, no API shape change.

## Scope note

Local Maven/Java is not installed in this workspace so the build wasn't run here. The diff is four lines of production code (one actual change + minor test-mock wiring) and the pattern is copied verbatim from the neighbouring `delete` endpoint, so the risk of a build regression is minimal -- but happy to iterate if CI surfaces anything.

Closes #862

---

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@
 
 - fix: add missing `@Transactional` to `@Modifying` queries in `ApiTokenRepository` and `SessionTokenRepository` to ensure proper transaction handling.
 - Removed unnecessary `@Transactional` annotations from methods in EssenciumScheduler.
+- `POST /v1/users/{id}/terminate` now enforces ownership-based access control (`testAccess(spec)`) before fetching the target user, matching the pattern already used by `update`, `patch`, and `delete`. Previously, any user with `USER_UPDATE` could terminate sessions of any other user even when `@RestrictAccessToOwnedEntities` was configured.
 
 ### 🔨 Dependency Upgrades
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,19 @@
 # Migration Guide
 
+## Version `3.4.0`
+
+### Access-control change: `terminate` endpoint honours ownership specs
+
+`POST /v1/users/{id}/terminate` now runs `testAccess(spec)` before fetching the
+target user. Downstream projects that rely on ownership-based access control
+(e.g. `@RestrictAccessToOwnedEntities`) will see `terminate` start respecting
+those restrictions; projects relying on `@Secured` only (`USER_UPDATE`) are
+unaffected because `testAccess` returns the service unchanged when no ownership
+filter applies. No database schema change. No API shape change. Clients that
+were previously allowed to terminate another user's session will now receive a
+`ResourceNotFoundException` — the same response `delete` / `update` would have
+returned under the same conditions before this change.
+
 ## Version `3.3.0`
 
 ### UserUtil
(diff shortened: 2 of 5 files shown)


pwannenmacher (changes requested):
Could you please rebase onto `main`?

tmchow:
Rebased onto main. Only conflict was the CHANGELOG bug-fix bullet list (combined cleanly with the new `@Transactional` fixes already in main). Code unchanged.

pwannenmacher (approved):

PR #894, opened 2026-04-23
817 upgrade to spring boot 40x and switch to multi module

pwannenmacher:


essencium-backend/pom.xml
```diff
@@ -5,16 +5,17 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-starter-parent</artifactId>
-        <version>4.1.0</version>
+        <groupId>de.frachtwerk</groupId>
+        <artifactId>essencium-backend-parent</artifactId>
```
jnsfZ, line 9: **Published library would reference a parent that is never published**

**Problem:** The library now inherits from `essencium-backend-parent`, but the release job only deploys `essencium-backend` (`mvn deploy -pl essencium-backend`, no `-am`) and there's no flatten step. So the POM uploaded to Maven Central still points at a parent that never gets uploaded, and downstream projects can't resolve it. (On `main` it inherits `spring-boot-starter-parent`, which is already on Central — so this is new.)

**Goal:** The artifact on Maven Central must be self-contained and usable by consumers.

**Recommendation:** Add `flatten-maven-plugin` (e.g. `flattenMode=bom`) to this module so the published POM has no unresolved parent. Note: just adding `-am` to the deploy step isn't enough — the parent POM would also need to be GPG-signed.

pom.xml
```diff
@@ -0,0 +1,392 @@
...
+                <version>${zxcvbn.version}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
```
jnsfZ: **`hibernate-processor` is silently dropped here**

**Problem:** `maven-compiler-plugin` is configured twice in this parent — here in `<build><plugins>`, and again below in `<pluginManagement>`. This active copy wins and leaves out `hibernate-processor`, so the JPA metamodel processor no longer runs (I confirmed the build generates zero `*_` classes; on `main` it does run). It's harmless today, but breaks as soon as any code uses a generated metamodel class.

**Goal:** One compiler configuration, behaving the same as `main`.

**Recommendation:** Remove this partial copy and keep only the full `<pluginManagement>` definition (the one that includes `hibernate-processor`).

pom.xml
```diff
@@ -0,0 +1,392 @@
...
+                    <artifactId>jacoco-maven-plugin</artifactId>
+                    <version>${jacoco.plugin.version}</version>
+                    <executions>
+                        <execution>
+                            <id>prepare-agent</id>
+                            <goals>
+                                <goal>prepare-agent</goal>
+                            </goals>
+                            <configuration>
+                                <destFile>${project.build.directory}/jacoco.exec</destFile>
```
jnsfZ, line 172: **This breaks the `sonarcloud.yml` coverage job**

**Problem:** Hard-coding the jacoco `destFile` here overrides the `-Djacoco.destFile=exportJacoco/...` that `sonarcloud.yml` still passes (that workflow wasn't migrated). The exec file ends up in `target/`, `exportJacoco/` is never created, and the workflow's `mv exportJacoco/* target/` then fails. I verified this.

**Goal:** Sonar coverage keeps working after the restructure.

**Recommendation:** Migrate `sonarcloud.yml` to the new layout — drop the `exportJacoco`/`-Djacoco.destFile` indirection and read `target/jacoco.exec` + `target/jacoco-it.exec` directly.

CHANGELOG.md
```diff
@@ -109,6 +112,10 @@
 - upgraded net.kaczmarzyk:specification-arg-resolver from 3.2.2 to 3.4.0
 - removed spring-boot-properties-migrator from dependencies
 
+- upgraded org.springframework.boot:spring-boot-starter-parent from 3.5.8 to 4.0.0
+- upgraded org.springdoc:springdoc-openapi-starter-webmvc-ui from 2.8.14 to 3.0.0
+- upgraded net.kaczmarzyk:specification-arg-resolver from 3.2.2 to 4.0.0
```
jnsfZ: **Wrong version numbers (duplicate block)**

**Problem:** These say `4.0.0 / 3.0.0 / 4.0.0`, but the actual versions are `4.1.0 / 3.0.3 / 4.1.0` — already listed correctly under `4.0.0-SNAPSHOT` (lines 17–19). They're also under the released `3.4.0` section.

**Goal:** No conflicting or incorrect version history.

**Recommendation:** Delete these three lines; the correct entries already exist above.

pom.xml
```diff
@@ -0,0 +1,392 @@
...
+            <dependency>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-lang3</artifactId>
+                <version>3.20.0</version>
+            </dependency>
+
+            <!-- essencium-backend-core dependency -->
+            <dependency>
+                <groupId>de.frachtwerk</groupId>
+                <artifactId>essencium-backend-core</artifactId>
```
jnsfZ: **Manages a module that doesn't exist**

**Problem:** `dependencyManagement` declares `essencium-backend-core`, but there's no such artifact. Meanwhile the dependency the demo apps actually use — `essencium-backend` — isn't managed, so both still hard-code `<version>${project.version}</version>`.

**Goal:** Centralised version management that matches the real modules.

**Recommendation:** Rename this entry to `essencium-backend` (then the demo modules can drop their explicit version), or remove it.

pom.xml
```diff
@@ -0,0 +1,392 @@
...
+        <maven-failsafe-plugin.version>3.5.5</maven-failsafe-plugin.version>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <!-- Override Spring Boot BOM to avoid vulnerable version of com.fasterxml.jackson.core:jackson-core (GHSA-72hv-8253-57qq) -->
+            <dependency>
+                <groupId>com.fasterxml.jackson</groupId>
+                <artifactId>jackson-bom</artifactId>
+                <version>2.21.2</version>
```
jnsfZ: **Unintended jackson downgrade**

**Problem:** `jackson-bom` goes `2.21.3 → 2.21.2` compared to `main`. Not a security issue (both versions are patched), it just looks accidental.

**Goal:** No accidental dependency drift.

**Recommendation:** Restore `2.21.3`, or confirm this override is still needed at all under Boot 4 / Jackson 3.

jnsfZ (changes requested):

pom.xml
```diff
@@ -0,0 +1,359 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
pwannenmacher: removed

jnsfZ (approved):

PR #897, opened 2026-05-04
fix: handle validation errors without arguments

### LINKED ISSUE #867: fix: RestExceptionHandler crashes on validation errors without arguments
## Problem

The `RestExceptionHandler` assumes every validation error has at least one argument. It accesses `getArguments()[0]` without null or bounds checks. When a validation error has no arguments (e.g. from custom validators), the handler itself crashes with `NullPointerException` or `ArrayIndexOutOfBoundsException`.

The result: instead of a helpful HTTP 400 response, the client gets a generic HTTP 500, and the original validation message is lost.

## Context

- **File:** [`RestExceptionHandler.java`](essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandler.java), lines 92-93

```java
// crashes if getArguments() is null or empty
DefaultMessageSourceResolvable messageSourceResolvableArgument =
    (DefaultMessageSourceResolvable) messageSourceResolvable.getArguments()[0];
```
- **Impact:** Affects all endpoints with validation. Especially relevant for custom validators (essencium defines `@StrongPassword` and `@ValidEmail`) that may not provide arguments in the expected format.
- **The problem is in the error handler itself:** Spring cannot handle the error gracefully when the handler crashes.

## Acceptance Criteria

- [ ] The handler processes validation errors correctly even when `getArguments()` is null or empty
- [ ] Clients always receive HTTP 400 with a readable error message for validation failures
- [ ] Test simulating a validation error without arguments
- [ ] `CHANGELOG.md` updated
- [ ] `MIGRATION.md` updated (if the fix changes behavior for downstream projects)

huyua9:
## Summary
- Guard validation error formatting when a resolvable error has null or empty arguments.
- Fall back to the validation error default message instead of crashing the exception handler.
- Add regression coverage for field arguments, empty arguments, and null arguments.
- Update the changelog.

Fixes #867.

## Tests
- `git diff --check`

Note: I could not run the Maven test suite locally because this machine does not have Maven installed and only has Java 8/10, while this project requires Java 21.

### DIFF
--- a/essencium-backend/src/test/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandlerTest.java
+++ b/essencium-backend/src/test/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandlerTest.java
@@ -28,7 +28,7 @@ class RestExceptionHandlerTest {
 
   @Test
   void formatValidationErrorUsesFieldArgumentWhenPresent() {
-    var fieldArgument = new DefaultMessageSourceResolvable("name");
+    var fieldArgument = new DefaultMessageSourceResolvable(new String[] {"name"}, "name");
     var resolvable =
         new DefaultMessageSourceResolvable(
             new String[] {"validation.name"}, new Object[] {fieldArgument}, "must not be blank");
--- a/essencium-backend/src/test/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandlerTest.java
+++ b/essencium-backend/src/test/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandlerTest.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2026 Frachtwerk GmbH, Leopoldstraße 7C, 76133 Karlsruhe.
+ *
+ * This file is part of essencium-backend.
+ *
+ * essencium-backend is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * essencium-backend is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with essencium-backend. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package de.frachtwerk.essencium.backend.controller.advice;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.context.support.DefaultMessageSourceResolvable;
+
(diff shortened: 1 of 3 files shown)


essencium-backend/src/test/java/de/frachtwerk/essencium/backend/controller/advice/RestExceptionHandlerTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.context.support.DefaultMessageSourceResolvable;
+
+class RestExceptionHandlerTest {
+
+  @Test
+  void formatValidationErrorUsesFieldArgumentWhenPresent() {
```
jnsfZ, line 30: Thanks for the fix! The production change in `RestExceptionHandler` is correct — the three crash scenarios from #867 (`null` args, empty args, wrong type) are all handled.

I ran the test suite locally though, and one of the new tests fails:

`RestExceptionHandlerTest.formatValidationErrorUsesFieldArgumentWhenPresent`

Expected:
```text
"name must not be blank"
```

But was:
```text
"must not be blank"
```

The cause is in the test fixture:

```java
var fieldArgument = new DefaultMessageSourceResolvable("name");
```

The `DefaultMessageSourceResolvable(String)` constructor sets the code to `"name"` and leaves `defaultMessage` as `null`.

So in the production code:

```java
messageSourceResolvableArgument.getDefaultMessage()
```

returns `null`, the new guard:

```java
if (field != null && !field.isBlank())
```

falls through, and the result is just the validation message without the field prefix.

That means the test currently passes through the same fallback branch as the other two tests, instead of exercising the `"field + message"` happy path it's named after.

Could you switch to a constructor that actually sets the default message? For example:

```java
var fieldArgument =
    new DefaultMessageSourceResolvable(new String[] {"name"}, "name");
```

After that change the assertion should pass and the test will cover the path it claims to cover.

jnsfZ (changes requested):

huyua9:
Thanks for catching that — I pushed `4983bec` with the fixture correction.

Change:
- Switched the field argument fixture to `new DefaultMessageSourceResolvable(new String[] {"name"}, "name")`, so `getDefaultMessage()` returns `"name"` and the test now exercises the intended `field + message` path.

Validation:
- I could not run the Maven test locally because this checkout has no `mvnw` and the local environment does not have `mvn` installed.
- I did re-check the change scope: this is test-only and matches the requested constructor fix directly.

huyua9:
Thanks for the review. The test fixture has been updated so `formatValidationErrorUsesFieldArgumentWhenPresent` now uses a `DefaultMessageSourceResolvable` constructor that sets the field default message to `name`, which exercises the intended `field + message` path.

Current head includes:
- `var fieldArgument = new DefaultMessageSourceResolvable(new String[] {"name"}, "name");`
- expected result remains `name must not be blank`

Validation from this checkout:
- `git diff --check origin/main...HEAD` passed
- conflict marker scan on the touched handler/test files passed
- targeted Maven test could not be run locally because this machine has no `mvn`/`mvnw` available and Java is 10 while the project requires a newer Java toolchain.

jnsfZ (approved):

PR #905, opened 2026-05-08
feat: implement IP whitelisting for API token usage

pwannenmacher:


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +132,34 @@ protected void successfulAuthentication(
...
+                        .collect(Collectors.toCollection(HashSet::new)))
+            .orElseGet(HashSet::new);
+
+    if (remoteAddr != null && !remoteAddr.isEmpty()) {
+      forwardedIpList.add(remoteAddr);
+    }
+
+    if (forwardedIpList.stream().noneMatch(this::isIpAllowed)) {
+      throw new NotAllowedException("IP address not allowed to use API tokens");
+    }
```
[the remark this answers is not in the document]
pwannenmacher: NotAllowedException from Essencium results in HTTP status 403.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -77,12 +88,26 @@ public Authentication attemptAuthentication(
                     new AuthenticationCredentialsNotFoundException(
                         "missing authorization header parameter"));
 
-    return getAuthentication(token);
+    return getAuthentication(token, request);
   }
 
-  public Authentication getAuthentication(String token) {
+  public Authentication getAuthentication(String token, HttpServletRequest request) {
```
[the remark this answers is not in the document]
pwannenmacher: Solved by CHANGELOG- and MIGRATION-Docs

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/JwtTokenService.java
```diff
@@ -273,15 +274,14 @@ private SessionToken createToken(
...
+  public Jws<Claims> verifyToken(String token) {
     try {
       return Jwts.parser()
           .keyLocator(sessionTokenKeyLocator)
           .requireIssuer(appJwtProperties.getIssuer())
           .clock(this)
           .build()
-          .parseSignedClaims(token)
-          .getPayload();
+          .parseSignedClaims(token);
```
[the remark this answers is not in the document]
pwannenmacher: Solved by CHANGELOG- and MIGRATION-Docs

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +138,78 @@ protected void successfulAuthentication(
...
+      if (!isProxyTrusted(ip)) {
+        return ip;
+      }
+    }
+
+    // All hops are trusted proxies — fall back to leftmost (closest to the original client)
+    return chain.getFirst();
+  }
+
+  private boolean isIpAllowed(String ip) {
```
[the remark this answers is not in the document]
pwannenmacher: Fully addressed; comment no longer relevant.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/model/validation/IpOrCidrValidator.java
```diff
@@ -0,0 +1,67 @@
...
+
+  private static final Pattern IPV4 =
+      Pattern.compile("^\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}(/\\d{1,3})?$");
+
+  @Override
+  public boolean isValid(String value, ConstraintValidatorContext context) {
+    if (value == null || value.isBlank()) {
+      return false;
+    }
+    String trimmed = value.trim();
```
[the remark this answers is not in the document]
pwannenmacher: Fully addressed; comment no longer relevant.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -77,12 +91,26 @@ public Authentication attemptAuthentication(
...
+
+      if (Objects.nonNull(appTokenProperties.getAllowedIpAddresses())
+          && !appTokenProperties.getAllowedIpAddresses().isEmpty()
+          && SessionTokenType.valueOf(type).equals(SessionTokenType.API)) {
+        verifyIpAddress(request);
+      }
+
+      if (Objects.nonNull(appTokenProperties.getPresharedSecrets())
+          && !appTokenProperties.getPresharedSecrets().isEmpty()
+          && SessionTokenType.valueOf(type).equals(SessionTokenType.API)) {
```
jnsfZ: **🔴 Critical — uncaught `NullPointerException` / `IllegalArgumentException` from `SessionTokenType.valueOf(type)`.**

A syntactically valid, Essencium-signed JWT can reach this block with a missing or unexpected `typ` header. `valueOf(...)` on lines 105 and 111 then crashes the filter chain — and neither exception is caught by the `try` block (lines 126–129 only catch `SessionAuthenticationException` / `SignatureException`). Result: the client sees a 500 instead of a clean 401/403.

### Why this can happen

- ✅ **`typ` is optional in the JWT spec.** RFC 7519 §5.1 defines `typ` as OPTIONAL, and JJWT's own README example shows a header without it (`{"alg": "HS256"}`). So `jws.getHeader().getType()` returning `null` is a normal outcome, not an attack — any externally-issued token signed with a key Essencium's `SessionTokenKeyLocator` happens to accept can land here without `typ`.
- ✅ **`Enum.valueOf` throws on `null` and on unknown names.** Per the JDK Javadoc (`java.lang.Enum#valueOf`): `NullPointerException` if the name is `null`; `IllegalArgumentException` if no enum constant matches.

### Suggested fix

Parse `type` once, defensively, and use the boolean in both branches (also removes the duplicated `valueOf` call and the redundant `Objects.nonNull` guards on the `@NotNull`-initialised sets):

```java
String type = jws.getHeader().getType();
boolean isApiToken = SessionTokenType.API.name().equals(type);

if (isApiToken && !appTokenProperties.getAllowedIpAddresses().isEmpty()) {
  verifyIpAddress(request);
}
if (isApiToken && !appTokenProperties.getPresharedSecrets().isEmpty()) {
  verifyPresharedSecret(request);
}
```

Please add a regression test for a token whose `typ` header is missing and one with an unrecognised value.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +141,81 @@ protected void successfulAuthentication(
...
+    if (!ipOrCidrValidator.isValid(ip, null)) {
+      return false;
+    }
+    return appTokenProperties.getTrustedProxies().stream()
+        .anyMatch(trusted -> new IpAddressMatcher(trusted.trim()).matches(ip));
+  }
+
+  private void verifyPresharedSecret(HttpServletRequest request) {
+    String header = request.getHeader(appTokenProperties.getPresharedSecretHeaderName());
+    if (header == null || !appTokenProperties.getPresharedSecrets().contains(header)) {
```
jnsfZ: **🟠 Major — non-constant-time PSK comparison (timing-attack surface).**

`Set.contains(...)` uses `String.equals`, which short-circuits on the first differing character. The PSK is a long-lived shared secret used to gate access to API tokens — treating it as a credential and comparing with non-constant-time equality is the textbook timing-attack pattern (✅ OWASP "Use of Hard-coded Password" / cheat-sheet guidance on secret comparison).

### Suggested fix

Iterate and use `MessageDigest.isEqual` on the UTF-8 byte arrays — same JDK, no extra dependency, constant-time over the configured secrets:

```java
private void verifyPresharedSecret(HttpServletRequest request) {
  String header = request.getHeader(appTokenProperties.getPresharedSecretHeaderName());
  if (header == null) {
    throw new NotAllowedException("Invalid preshared secret");
  }
  byte[] headerBytes = header.getBytes(StandardCharsets.UTF_8);
  boolean match = false;
  for (String secret : appTokenProperties.getPresharedSecrets()) {
    if (MessageDigest.isEqual(headerBytes, secret.getBytes(StandardCharsets.UTF_8))) {
      match = true;
      // intentionally no break — keep timing independent of the matching position
    }
  }
  if (!match) {
    throw new NotAllowedException("Invalid preshared secret");
  }
}
```

Minor side-note while here: the header value is not trimmed, so a trailing space on the client side silently fails the match — possibly intentional (strict), worth a deliberate decision.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/security/AppTokenProperties.java
```diff
@@ -0,0 +1,82 @@
...
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.validation.annotation.Validated;
+
+@Data
+@EqualsAndHashCode(callSuper = false)
+@Configuration
+@ConfigurationProperties(prefix = "app.token")
```
jnsfZ: **🟠 Major — property namespace is inconsistent with the rest of the project.**

All other auth-related properties live under `app.auth.*` — confirmed by grep:

- `AppJwtProperties` → `app.auth.jwt`
- `AppLdapProperties` → `app.auth.ldap`
- `AppOAuth2Properties` → `app.auth.oauth`

This bean uses `app.token`, which makes the API-token config harder to discover for operators who grep / scan `app.auth`. Renaming after release is a breaking change for downstream `application.yml` files, so it's worth getting right now.

**Suggested:** `app.auth.api-token` (or nest under jwt as `app.auth.jwt.api-token.*`).

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +141,81 @@ protected void successfulAuthentication(
     chain.doFilter(request, response);
   }
 
+  private void verifyIpAddress(HttpServletRequest request) {
```
jnsfZ, line 212: **🟠 Major — bypass risk if a downstream filter chain replaces or subclasses `JwtTokenAuthenticationFilter`.**

The IP and PSK checks live entirely inside `getAuthentication`. Because this is a library, downstream projects can — and sometimes do — register their own `AbstractAuthenticationProcessingFilter` for token paths, or subclass and override. In either case, the new gating is silently dropped without any compile-time or runtime signal.

### Suggested mitigations (pick one)

- Extract the IP / PSK check into a small, separately-injectable component (e.g. `ApiTokenAccessVerifier`) that downstream filters can keep wiring up explicitly.
- Or, at minimum, document this in `docs/pages/devguide/started/features/api-token-ip-whitelist.mdx`: "If you replace `JwtTokenAuthenticationFilter` in your application, re-invoke the verifier yourself."

No doc currently warns about this.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +141,81 @@ protected void successfulAuthentication(
...
+        .anyMatch(allowed -> new IpAddressMatcher(allowed.trim()).matches(ip));
+  }
+
+  private boolean isProxyTrusted(String ip) {
+    if (!ipOrCidrValidator.isValid(ip, null)) {
+      return false;
+    }
+    return appTokenProperties.getTrustedProxies().stream()
+        .anyMatch(trusted -> new IpAddressMatcher(trusted.trim()).matches(ip));
+  }
```
jnsfZ, line 283: **🟡 Minor — `IpAddressMatcher` allocated on the hot path, per request, per entry.**

`isIpAllowed` and `isProxyTrusted` both `new IpAddressMatcher(...)` inside the stream. For each authenticated API request, that's `N + M` instances (N = allowed addresses, M = trusted proxies in the chain). `IpAddressMatcher`'s constructor delegates to `InetAddressMatchers.fromIpAddress(...)` which parses the address each time.

Not a correctness bug, but trivially avoidable: cache the matchers once when properties are loaded (e.g. via `@PostConstruct` on `AppTokenProperties`, storing `List<IpAddressMatcher>` alongside the `Set<String>`), or memoise inside the filter.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +141,81 @@ protected void successfulAuthentication(
...
+
+  private void verifyPresharedSecret(HttpServletRequest request) {
+    String header = request.getHeader(appTokenProperties.getPresharedSecretHeaderName());
```
jnsfZ, line 286: **🟡 Minor — header value is compared verbatim, no trimming.**

`request.getHeader(...)` returns the raw header value. If a client (or intermediate proxy) inserts a stray trailing or leading space, the strict `Set.contains` lookup misses and the request is rejected as 403. Often the "correct" behaviour for security headers (be strict), but worth a deliberate decision and a comment so it isn't accidentally relaxed later.

(See the M2 PSK comment for the constant-time-comparison fix that should also land here — the two changes are in the same method.)

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -55,6 +65,10 @@ public class JwtTokenAuthenticationFilter<ID extends Serializable>
 
   @Autowired private JwtTokenService jwtTokenService;
 
+  @Autowired private AppTokenProperties appTokenProperties;
+
+  private final IpOrCidrValidator ipOrCidrValidator = new IpOrCidrValidator();
```
jnsfZ, line 81: **🟡 Minor — `IpOrCidrValidator` instantiated as a `new …()` field, bypassing Spring's validator wiring.**

The validator is stateless (good), and instantiating it directly works fine. But it's unusual compared to the rest of the codebase, which uses `@Autowired` for everything else here. Two options:

- Mark `static final` (it has no instance state) so it's at least a singleton.
- Or convert to a Spring bean and `@Autowired` it like `jwtTokenService` / `appTokenProperties`.

Very minor — purely a code-style nit.

jnsfZ (changes requested):

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/security/AppTokenProperties.java
pwannenmacher: done. see https://github.com/Frachtwerk/essencium-backend/pull/905/commits/4df2cfde8925f5837ed14d823f9152c096688b38

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
pwannenmacher: done. see https://github.com/Frachtwerk/essencium-backend/pull/905/commits/806e8dc26d3cb20e1bd8fc00a429635402abb62e

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
pwannenmacher: done. see https://github.com/Frachtwerk/essencium-backend/pull/905/commits/1e38a6291f159a875c9469c93461ed5fed6cc3a3

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
pwannenmacher: done. see https://github.com/Frachtwerk/essencium-backend/pull/905/commits/abd697b977a44d26c4643914e6d2cc897f8d6cf8

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/security/JwtTokenAuthenticationFilter.java
```diff
@@ -113,6 +149,98 @@ protected void successfulAuthentication(
...
+
+    if (StringUtils.isBlank(remoteAddr)) {
+      throw new NotAllowedException("Unable to determine remote IP address");
+    }
+
+    final String clientIp =
+        resolveClientIp(remoteAddr.trim(), request.getHeader("X-Forwarded-For"));
+
+    if (!ipOrCidrValidator.isValid(clientIp, null) || !isIpAllowed(clientIp)) {
+      throw new NotAllowedException("IP address not allowed to use API tokens");
```
jnsfZ, line 223: ## 🔴 Critical — rejection returns HTTP **500**, not the documented 403

Every `throw new NotAllowedException(...)` in this filter (lines 156, 163, 228, 240) actually produces an HTTP 500 with a leaked stack trace, not the 403 the docs and `CHANGELOG.md` promise. Repro from the dev app with PSK enabled:

```
HTTP/1.1 500
{"status":500,"error":"Internal Server Error",
 "message":"Invalid preshared secret",
 "trace":"de.frachtwerk.essencium.backend.model.exception.NotAllowedException...\n\tat ...JwtTokenAuthenticationFilter.verifyPresharedSecret(...)\n\tat ...attemptAuthentication(...)"}
```

Same for the IP whitelist path — same 500, same stack-trace leak, just a different message.

### Why this happens

`NotAllowedException` has `@ResponseStatus(HttpStatus.FORBIDDEN)`. That annotation is honoured by **Spring MVC's** `ResponseStatusExceptionResolver`, which runs only for exceptions that bubble up out of controller handlers. The Spring Security filter chain runs **before** that — so when the filter throws, the exception never reaches the MVC layer. The servlet container catches it and renders the default 500 error page.

Spring Security's own escape hatch is `ExceptionTranslationFilter`, which only translates two specific exception types:

> ✅ Doc-confirmed ([Spring Security reference, *Servlet Architecture → Handling Security Exceptions*](https://docs.spring.io/spring-security/reference/servlet/architecture.html#servlet-exceptiontranslationfilter)):
>
> ```java
> try {
>   filterChain.doFilter(request, response);
> } catch (AccessDeniedException | AuthenticationException ex) {
>   // ...
> }
> ```
>
> *"The `ExceptionTranslationFilter` allows for the translation of `AccessDeniedException` and `AuthenticationException` into HTTP responses."*

A custom `RuntimeException` like `NotAllowedException` matches neither, so it slips past untranslated. And there's a positional wrinkle on top: `JwtTokenAuthenticationFilter` is registered via `addFilterBefore(..., AnonymousAuthenticationFilter.class)` in `WebSecurityConfig`, which puts it **upstream** of `ExceptionTranslationFilter` in the chain — meaning even `AccessDeniedException` from this position wouldn't be caught (`ExceptionTranslationFilter` only wraps *downstream* `chain.doFilter` in its try/catch).

### Impact

Clients written against the documented `403` contract will mis-classify the response (500 = transient server problem → retry; 403 = config problem → don't retry), and a stack trace ends up in the body.

### Minimal fix

`AbstractAuthenticationProcessingFilter` (the parent class) already invokes `unsuccessfulAuthentication` whenever `attemptAuthentication` throws an `AuthenticationException`. That's the right hook — it runs after the security chain has unwound to the failure point, so `response.sendError(...)` is safe there (Spring Security uses the same pattern in `Http403ForbiddenEntryPoint`).

The existing filter already uses `BadCredentialsException` (an `AuthenticationException`) to signal token-verification failures at lines 134–139. Adding a sibling exception for the 403 case follows the established pattern in the same class.

**One new exception class:**

```java
// new file: model/exception/ApiTokenAccessRejectedException.java
public class ApiTokenAccessRejectedException extends AuthenticationException {
  public ApiTokenAccessRejectedException(String message) {
    super(message);
  }
}
```

**Change 4 throw sites in this file** (lines 156, 163, 228, 240) from:

```java
throw new NotAllowedException("IP address not allowed to use API tokens");
```

to:

```java
throw new ApiTokenAccessRejectedException("IP address not allowed to use API tokens");
```

**Override `unsuccessfulAuthentication`** to render 403 for this exception type:

```java
@Override
protected void unsuccessfulAuthentication(
    HttpServletRequest request, HttpServletResponse response, AuthenticationException failed)
    throws IOException, ServletException {
  if (failed instanceof ApiTokenAccessRejectedException) {
    response.sendError(HttpStatus.FORBIDDEN.value(), failed.getMessage());
    return;
  }
  super.unsuccessfulAuthentication(request, response, failed);
}
```

**Diff total**: ~12 lines in `JwtTokenAuthenticationFilter` + one new ~5-line exception class. Zero changes to `WebSecurityConfig`, zero new beans, zero `@ControllerAdvice` edits. The existing `RestExceptionHandler.handleNotAllowedException` keeps working for any **controller-thrown** `NotAllowedException` elsewhere in the codebase (unchanged).

### Why `sendError` is OK in this position

Generally `sendError` from the middle of a security filter is bad — it triggers the servlet error-page dispatcher and re-enters the filter chain. **From inside `unsuccessfulAuthentication`** it's the conventional pattern: the chain has already unwound to the failure-handling point, no re-entry happens, and `BasicErrorController` renders a JSON body with the same `{timestamp, status, error, message, path}` shape the rest of this API already returns. In production where `server.error.include-message=never` (the Spring default), the message is suppressed — which is arguably *better* security than the current behaviour, since attackers don't learn whether their PSK was wrong or their IP was wrong.

### Test case (fails on current code, passes after the fix)

Drop into `JwtTokenAuthenticationFilterTest` — asserts on the actual HTTP status the client sees:

```java
@Test
void deniedPsk_writes403_notInternalServerError() throws Exception {
  when(jwtTokenService.verifyToken(anyString())).thenReturn(jws);
  when(jws.getHeader()).thenReturn(jwsHeader);
  when(jws.getPayload()).thenReturn(claims);
  when(jwsHeader.getType()).thenReturn(SessionTokenType.API.name());
  when(claims.entrySet()).thenReturn(Set.of());
  when(claims.get(JwtTokenService.CLAIM_ROLES)).thenReturn(List.of());
  when(claims.get(JwtTokenService.CLAIM_RIGHTS)).thenReturn(List.of());

  when(appTokenProperties.getAllowedIpAddresses()).thenReturn(Set.of());
  when(appTokenProperties.getPresharedSecrets()).thenReturn(Set.of("right-secret"));
  when(appTokenProperties.getPresharedSecretHeaderName()).thenReturn("X-API-Token-PSK");

  MockHttpServletRequest request = new MockHttpServletRequest();
  request.addHeader(HttpHeaders.AUTHORIZATION, "Bearer aaa.bbb.ccc");
  // no X-API-Token-PSK header
  MockHttpServletResponse response = new MockHttpServletResponse();
  FilterChain chain = mock(FilterChain.class);

  filter.doFilter(request, response, chain);

  // Currently FAILS: filter throws NotAllowedException which bubbles up as 500.
  assertThat(response.getStatus()).isEqualTo(HttpStatus.FORBIDDEN.value());
  verify(chain, never()).doFilter(any(), any());
}
```

And the matching test for the IP-whitelist throw site:

```java
@Test
void deniedIp_writes403_notInternalServerError() throws Exception {
  when(jwtTokenService.verifyToken(anyString())).thenReturn(jws);
  when(jws.getHeader()).thenReturn(jwsHeader);
  when(jws.getPayload()).thenReturn(claims);
  when(jwsHeader.getType()).thenReturn(SessionTokenType.API.name());
  when(claims.entrySet()).thenReturn(Set.of());
  when(claims.get(JwtTokenService.CLAIM_ROLES)).thenReturn(List.of());
  when(claims.get(JwtTokenService.CLAIM_RIGHTS)).thenReturn(List.of());

  when(appTokenProperties.getAllowedIpAddresses()).thenReturn(Set.of("1.2.3.4"));
  when(appTokenProperties.getTrustedProxies()).thenReturn(Set.of());
  when(appTokenProperties.getPresharedSecrets()).thenReturn(Set.of());

  MockHttpServletRequest request = new MockHttpServletRequest();
  request.setRemoteAddr("9.9.9.9");
  request.addHeader(HttpHeaders.AUTHORIZATION, "Bearer aaa.bbb.ccc");
  MockHttpServletResponse response = new MockHttpServletResponse();
  FilterChain chain = mock(FilterChain.class);

  filter.doFilter(request, response, chain);

  assertThat(response.getStatus()).isEqualTo(HttpStatus.FORBIDDEN.value());
  verify(chain, never()).doFilter(any(), any());
}
```

Drop the PSK test into the existing `PresharedSecret` `@Nested` class, the IP test into `IpWhitelisting` — same place their happy-path counterparts already live.

docs/pages/devguide/started/features/api-token-ip-whitelist.mdx
```diff
@@ -0,0 +1,178 @@
...
+    token:
+      allowed-ip-addresses:
+        - 203.0.113.42 # gateway or partner system to allow
+      trusted-proxies:
+        - 172.18.0.0/16 # internal Docker network (proxy containers)
+        - fd11::/16 # IPv6 range used by some proxy images (e.g. nginx, envoy)
+```
+
+<Callout type="warning">
+  `allowed-ip-addresses` and `trusted-proxies` must be **disjoint**. An address
```
jnsfZ, line 42: Dont think this is Major, maybe worth looking at it @pwannenmacher 


## 🟠 Major — dual-stack pitfall: `trusted-proxies` (and the allowlist) need both IPv4 and IPv6 forms

When I ran the dev app locally on Linux and connected from `127.0.0.1`, `HttpServletRequest.getRemoteAddr()` consistently returned `::1` (IPv6 loopback), not `127.0.0.1`. With only the IPv4 form of the proxy listed in `trusted-proxies`, the whole XFF resolution stops working — even `curl --resolve localhost:8098:127.0.0.1` (forcing IPv4 from the client side) didn't change the picture.

### Repro from the dev app

```yaml
app:
  auth:
    token:
      allowed-ip-addresses:
        - 192.0.2.99
      trusted-proxies:
        - 127.0.0.1   # IPv4 only — looks correct, isn't
```

```
$ curl -H "Authorization: Bearer $API_TOKEN" -H "X-Forwarded-For: 192.0.2.99" \
       http://localhost:8098/v1/users
HTTP/1.1 500
{"message":"IP address not allowed to use API tokens"}
```

Add `::1` to `trusted-proxies` and the same request succeeds.

### Why this happens

On a default Linux JVM, `java.net.preferIPv4Stack=false`, so the embedded Tomcat binds to `[::]` and accepts both IPv4 and IPv6 connections. Loopback connections frequently arrive as `::1` even when the client *tried* to use `127.0.0.1`. The PR's `IpAddressMatcher` correctly treats them as different — ✅ doc-confirmed (Spring Security Javadoc for `IpAddressMatcher`):

> *"Both IPv6 and IPv4 addresses are supported, but a matcher which is configured with an IPv4 address will never match a request which returns an IPv6 address, and vice-versa."*

So `IpAddressMatcher("127.0.0.1").matches("::1")` returns `false` — the matcher is doing its job, the configuration is just incomplete. The resolution algorithm then walks past the (untrusted-looking) `::1`, treats it as the client IP, and rejects it.

Same trap exists for any Docker / Kubernetes setup where the proxy can connect over either stack — operators who copy the existing `172.18.0.0/16` example and forget the matching IPv6 range (or vice versa) will silently lock themselves out.

### Minimal fix

Docs-only callout in this MDX file. No code change. The user-visible problem is *silent lockout because the trap is undocumented*; once the docs name the trap, the operator knows what to configure.

Add this `<Callout>` right after the existing disjointness one (around line 47):

```mdx
<Callout type="warning">
  **Dual-stack note.** On hosts where the JVM listens on both IPv4 and IPv6
  (the Linux default with `java.net.preferIPv4Stack=false`), the same proxy
  can present its client address as either `192.0.2.1` or `::ffff:192.0.2.1`,
  and a `localhost` connection often arrives as `::1` rather than `127.0.0.1`.
  Spring Security's matcher treats the families as disjoint, so list **both**
  forms in `trusted-proxies` (and `allowed-ip-addresses` where relevant) for
  proxies and clients that could reach the application over either stack.
</Callout>
```

### Test case (encodes the gotcha so future readers see why both forms matter)

Add to `JwtTokenAuthenticationFilterTest.ResolveClientIp.cases()`. Both cases pass on current code — they're documentation-as-test:

```java
Arguments.of(
    "IPv4-only trusted-proxies vs IPv6 loopback: XFF is NOT honoured",
    Set.of("127.0.0.1"),       // trustedProxies (IPv4 only)
    "::1",                      // remoteAddr (dual-stack default)
    "192.0.2.99",               // XFF claims the real client
    "::1"),                     // resolved IP = ::1, NOT 192.0.2.99
Arguments.of(
    "Both v4 + v6 listed: XFF is honoured",
    Set.of("127.0.0.1", "::1"),
    "::1",
    "192.0.2.99",
    "192.0.2.99")              // real client correctly resolved
```

jnsfZ (changes requested):

jnsfZ (approved):

PR #929, opened 2026-06-24
918 exceptionhandling rebased

### LINKED ISSUE #918: ExceptionHandling

**Additional context**

jnsfZ:


AndriiFW:
Implemented and verified the new RFC 9457 `ProblemDetail` error handling.

### What was changed

- Added centralized error handling via `GlobalExceptionHandler`.
- Added `ProblemDetailFactory` to create consistent `application/problem+json` responses.
- Replaced legacy `ErrorResponse` / `RestExceptionHandler` handling.
- Removed `@ResponseStatus` from domain exceptions so HTTP status mapping is handled centrally.
- Added structured error codes for `ProblemDetail.type`, for example:
  - `urn:frachtwerk:error:NOT_FOUND`
  - `urn:frachtwerk:error:VALIDATION_FAILED`
  - `urn:frachtwerk:error:UNIQUE_CONSTRAINT_VIOLATION`
- Added configurable URN prefix via `essencium.error.urn-prefix`.
- Added `timestamp`, optional `stackTrace`, and validation `fieldErrors` as ProblemDetail extension properties.
- Added fallback handling for `detail`, so responses always contain a safe detail value even when an exception message is missing.

### Error mappings covered

The mapping table is covered by tests, including:

- `ResourceNotFoundException` → `404 NOT_FOUND`
- `InvalidInputException` / `ResourceUpdateException` → `400 INVALID_INPUT`
- `DuplicateResourceException` → `409 DUPLICATE_RESOURCE`
- `NotAllowedException` → `403 FORBIDDEN`
- `TokenInvalidationException` → `401 TOKEN_INVALIDATION`
- `TranslationFileException` → `500 TRANSLATION_FILE_ERROR`
- Validation exceptions → `400 VALIDATION_FAILED` with `fieldErrors`
- Malformed JSON → `400 MALFORMED_REQUEST`
- PostgreSQL SQL states:
  - `23505` → `409 UNIQUE_CONSTRAINT_VIOLATION`
  - `23503` → `409 FOREIGN_KEY_VIOLATION`
  - `23502` → `400 NOT_NULL_VIOLATION`

### Manual verification

Manually tested via Postman:

- `NOT_FOUND`
- `MALFORMED_REQUEST`
- `VALIDATION_FAILED` with `fieldErrors`
- `FORBIDDEN`
- `DUPLICATE_RESOURCE`
- real PostgreSQL `23505` unique constraint violation via duplicate API token description
- `include-message=never` sanitizes `detail`
- `include-stacktrace=never`, `on_param`, and `always`
- custom `essencium.error.urn-prefix`

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,241 @@
...
+
+    ProblemDetail problemDetail =
+        problemDetailFactory.create(
+            HttpStatus.BAD_REQUEST,
+            ErrorCode.VALIDATION_FAILED,
+            "Validation failed",
+            exception,
+            request);
+
+    problemDetail.setProperty("fieldErrors", fieldErrors);
```
jnsfZ: **Field errors are returned even when messages are suppressed**

**Problem:** `fieldErrors` is set unconditionally, while `detail` is hidden when `include-message=never` (same on line 158). So in a production-safe config the response still exposes the field names and messages.
**Why it matters:** it partly defeats turning messages off, and it's inconsistent — one setting is expected to control what the API exposes.
**Recommendation:** decide whether field errors should be visible in production; if not, gate them behind the same setting (centralised in `ProblemDetailFactory`). If yes, note it as intentional.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,241 @@
...
+      TokenInvalidationException exception, HttpServletRequest request) {
+    return createResponse(
+        HttpStatus.UNAUTHORIZED,
+        ErrorCode.TOKEN_INVALIDATION,
+        exception.getMessage(),
+        exception,
+        request);
+  }
+
+  @ExceptionHandler(TranslationFileException.class)
```
jnsfZ, line 125: **This 5xx path isn't logged**

**Problem:** the handler logs nothing, so a handled 500 returns to the client but leaves no server-side trace; with stack traces off (the default) the cause is lost.
**Why it matters:** the 500 becomes hard to diagnose — and less traceable than an unmapped error, which the framework still logs.
**Recommendation:** log 5xx at ERROR (with the exception) before returning.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,241 @@
...
+        exception,
+        request);
+  }
+
+  @ExceptionHandler(DataIntegrityViolationException.class)
+  public ResponseEntity<ProblemDetail> handleDataIntegrityViolationException(
+      DataIntegrityViolationException exception, HttpServletRequest request) {
+    String sqlState = findSqlState(exception);
+
+    if ("23505".equals(sqlState)) {
```
jnsfZ: **Avoid magic strings for the SQL-state codes**

**Problem:** `23505` / `23503` / `23502` are magic literals inline.
**Why it matters:** their meaning isn't obvious at a glance and they can't be referenced/tested by name.
**Recommendation:** extract to named constants — these are PostgreSQL/ANSI SQLSTATE codes, e.g. `SQLSTATE_UNIQUE_VIOLATION = "23505"`, `SQLSTATE_FOREIGN_KEY_VIOLATION = "23503"`, `SQLSTATE_NOT_NULL_VIOLATION = "23502"` — and switch on those.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,241 @@
...
+    if ("23502".equals(sqlState)) {
+      return createResponse(
+          HttpStatus.BAD_REQUEST,
+          ErrorCode.NOT_NULL_VIOLATION,
+          "Not null violation",
+          exception,
+          request);
+    }
+
+    throw exception;
```
jnsfZ: **Unmapped DB errors drop the `problem+json` format**

**Problem:** an unknown SQL state rethrows here, so the response falls back to Spring's default `application/json` — with no `type`/`title`/`detail`.
**Why it matters:** clients that parse the standard error shape don't get it on this path.
**Recommendation:** intentional per the spec (no catch-all), so a one-line comment here would help; optionally add a logged generic fallback so the contract still holds.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
```diff
@@ -0,0 +1,116 @@
...
+      HttpStatus status,
+      ErrorCode errorCode,
+      String detail,
+      Throwable throwable,
+      HttpServletRequest request) {
+    String resolvedDetail = resolveDetail(detail, request);
+
+    ProblemDetail problemDetail = ProblemDetail.forStatus(status);
+    problemDetail.setDetail(resolvedDetail);
+    problemDetail.setType(URI.create(errorProperties.getUrnPrefix() + errorCode.name()));
```
jnsfZ: **Error code is derived from the enum name — renames break clients**

**Problem:** `type` is built from `errorCode.name()`, so the public error code *is* the Java constant name.
**Why it matters:** renaming a constant (looks like an internal change) silently changes the public `type` that downstream clients match on.
**Recommendation:** give each `ErrorCode` an explicit wire value (e.g. `NOT_FOUND("NOT_FOUND")`) and read that — or add a "do not rename" note on the enum.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
```diff
@@ -0,0 +1,116 @@
...
+    problemDetail.setTitle(status.getReasonPhrase());
```
jnsfZ: **Redundant `setTitle` (minor)**

`ProblemDetail.getTitle()` already derives the status reason phrase when no title is set, so this line has no observable effect and can be removed.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/EssenciumErrorProperties.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package de.frachtwerk.essencium.backend.configuration.properties;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "essencium.error")
+public class EssenciumErrorProperties {
+
+  private String urnPrefix = "urn:frachtwerk:error:";
```
jnsfZ: **Validate the URN prefix**

**Problem:** `urnPrefix` isn't validated. A blank value yields a `type` with no namespace; a value with an illegal URI character makes `URI.create(...)` throw *inside* the error handler.
**Why it matters:** a misconfiguration produces a broken contract or an error while building the error response.
**Recommendation:** add `@Validated` + `@NotBlank` (ideally a format check), matching the other config classes.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/FieldErrorResponse.java
```diff
@@ -0,0 +1,22 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with essencium-backend. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package de.frachtwerk.essencium.backend.controller.advice;
+
+public record FieldErrorResponse(String field, String message) {}
```
jnsfZ, line 25: **Guard against null field/message**

**Problem:** both components can be null (the validation message and the resolved parameter name are nullable), so an entry can end up `{ "field": null, "message": null }`.
**Recommendation:** normalise in a compact constructor (default a blank field/message to a placeholder) — don't throw, a 400 handler must not itself 500.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
AndriiFW: Adjusted. Field errors now follow the message visibility setting.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
AndriiFW: Fixed. Added ERROR logging for the handled 5xx case.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
AndriiFW: Fixed. Extracted the SQLSTATE values into named constants.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
AndriiFW: Added logged ProblemDetail fallback.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
AndriiFW: Added explicit wire values.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
AndriiFW: Removed

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/EssenciumErrorProperties.java
AndriiFW: Fixed. Added validation to EssenciumErrorProperties so the URN prefix cannot be blank.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/FieldErrorResponse.java
AndriiFW: Fixed. 
I used `@JsonInclude(Include.NON_NULL)` and covered the case with a small serialization test.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/EssenciumErrorProperties.java
```diff
@@ -0,0 +1,39 @@
...
+  @NotBlank private String urnPrefix = "urn:frachtwerk:error:";
+
+  public String getUrnPrefix() {
+    return urnPrefix;
+  }
+
+  public void setUrnPrefix(String urnPrefix) {
+    this.urnPrefix = urnPrefix;
+  }
+}
```
pwannenmacher, line 33: Feel free to use Lombok’s `@Data`annotation here. This annotation eliminates the need for getters and setters.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
pwannenmacher, line 1: In order to ensure that all exceptions are ultimately represented identically to the outside world, there should also be an ExceptionHandler for generic exceptions that are not defined in more detail.

pwannenmacher (changes requested):

pwannenmacher (changes requested):
And please do a clean rebase onto the current branch.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ErrorCode.java
pwannenmacher, line 1: If you make `ErrorCode` inherit from a suitable interface, and the `createResponse` method in the `GlobalExceptionHandler` expects this interface rather than an `ErrorCode`, developers can implement their own additional ‘`ErrorCodes`’ at the application level and extend the `GlobalExceptionHandler`. To do this, however, the `createResponse` method would need to be at least protected or even public.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -0,0 +1,257 @@
...
+  @ExceptionHandler(TokenInvalidationException.class)
+  public ResponseEntity<ProblemDetail> handleTokenInvalidationException(
+      TokenInvalidationException exception, HttpServletRequest request) {
+    return createResponse(
+        HttpStatus.UNAUTHORIZED,
+        ErrorCode.TOKEN_INVALIDATION,
+        exception.getMessage(),
+        exception,
+        request);
+  }
```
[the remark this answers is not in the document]
AndriiFW: Adjusted. TokenInvalidationException is used for internal failures during token invalidation, not for invalid credentials. The handler now returns 500 and logs the exception.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
AndriiFW: I noticed that the [ticket](https://github.com/Frachtwerk/essencium-backend-issues-intern/issues/1) explicitly states that exceptions not listed in the table should fall back to the Spring default and that there should be no custom catch-all.

When adding a generic Exception handler, some existing integration tests start returning 500 instead of the expected 400/401 because exceptions previously handled by Spring are now caught by the generic handler.

Should I still add the generic catch-all and explicitly preserve those Spring/Security cases, or should we keep the original ticket behavior without a catch-all?

pwannenmacher (approved):

PR #959, opened 2026-07-28
feat: enhance configuration properties documentation

pwannenmacher:


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/MailProperties.java
```diff
@@ -137,15 +238,30 @@ public NewUserMail getNewUserMailConfig() {
...
     private boolean enabled = true;
 
+    /**
+     * Translation key for the mail subject. The {@code ^[^$].*} pattern forbids a leading {@code $}
+     * (which would be interpreted as a raw/placeholder value rather than a translation key).
+     */
     @Pattern(regexp = "^[^$].*")
     private String subjectKey;
 
+    /** Name of the FreeMarker template used to render the mail. Mandatory ({@link NotEmpty}). */
```
jnsfZ: "Mandatory (`{@link NotEmpty}`)" is not currently true, and this wording appears eight times across the PR.

`MailProperties` does carry `@Validated` (`:53`), but the nested fields (`newUserMail`, `resetTokenMail`,
`newLoginMail`, `contactMail`, `smtp`, `branding`, `defaultSender`) have **no `@Valid`**, so validation is not
cascaded into these inner classes. A search for `@Valid` across `configuration/properties/` returns no results.
The constraints on `template` and `resetLink` therefore never fire; a missing template surfaces later as a
rendering failure.

Two options:

- add `@Valid` to the nested fields so the constraints actually apply (a behaviour change — probably better as its
  own PR), or
- reword to "required for this mail type to work (not validated at start-up)".

Independently of that: the annotation is visible one line below, so repeating it in prose adds little and is one
more thing to keep in sync. I would drop the parenthesis and keep only statements the annotation cannot make — for
example "Changing it invalidates previously issued tokens" in `AppJwtProperties`, which is genuinely useful.

jnsfZ (approved):

PR #964, opened 2026-07-28
fix: filter JWT reserved claims to prevent overwriting expiration tim…

### LINKED ISSUE #963: API-Token - `exp` from User-Token overrides `exp` in Api-Token

pwannenmacher:
…e in ApiTokenService


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
...
 - Made all mail templates (`ResetTokenMessage`, `NewUserMessage`, `NewLoginMessage`, `ContactMessage`, `Header`, `Footer`) null-safe by adding FreeMarker default operators (`!`) to every `${…}` interpolation. Previously a missing/null branding value threw a `freemarker.core.InvalidReferenceException`, which — because mail delivery is `@Async` and the send methods catch `TemplateException` — was only logged and swallowed, so the mail silently never arrived. This notably affected the reset-token and new-user mails, whose templates were the only ones dereferencing `${mailBranding.textColor}`; the login mail rendered fine because it never touches that field. Colors fall back to sensible defaults (`#00b5d6` / `#ffffff`) to keep the inline CSS valid.
+- API-Token - exp from User-Token overrides exp in Api-Token. JWT_RESERVED_CLAIMS are now filtered out of the additional claims in ApiTokenService to avoid overwriting the exp claim from the user token.
```
jnsfZ: The entry names `ApiTokenService`, but the filtering happens in `JwtTokenService`. It also covers `exp` only: `sub`, `iss` and `iat` are dropped as well, and `sub` changes what an API token says it belongs to. It carries the token's own name instead of the creating user's e-mail address, and since the authentication filter and the auditing configuration both take the user name from `sub`, `createdBy` and `updatedBy` change with it.

```suggestion
- Fixed API tokens inheriting the `exp` of the user token they were created with, which made them expire within the access token lifetime instead of at their configured `validUntil`. Claim names registered by RFC 7519 (`iss`, `sub`, `aud`, `exp`, `nbf`, `iat`, `jti`) are excluded when `additionalClaims` are added to a token, so they keep the values set by the token service. `sub` was affected in the same way, so an API token carries its own username (`<user>-api-token-<id>`) instead of the creating user's e-mail address; as `sub` is used as the authenticated principal, this also changes `createdBy` and `updatedBy` for entities written through an API token and is a ⚠️ breaking change ⚠️ for downstream projects matching on those columns. See `MIGRATION.md` for details.
```

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/JwtTokenService.java
```diff
@@ -82,6 +83,17 @@ public class JwtTokenService implements Clock {
...
+  private static final Set<String> JWT_RESERVED_CLAIMS =
+      Set.of(
+          "iss", // Issuer
+          "sub", // Subject
+          "aud", // Audience
+          "exp", // Expiration Time
+          "nbf", // Not Before
+          "iat", // Issued At
+          "jti" // JWT ID
+          );
```
jnsfZ: `io.jsonwebtoken.Claims` already declares these seven as constants, and `Claims` is imported here anyway, so the literals can drift from what the library actually looks for. The reason for the set is also nowhere recorded: the library documents nothing about a claim added by name replacing one set through a dedicated method, and nothing fails loudly if the filter goes away.

```suggestion
  // The JWT builder keeps registered and custom claims in a single map, so a claim taken from
  // additionalClaims replaces any value the builder itself has set for the same name. Claim names
  // registered by RFC 7519 (section 4.1) are therefore never copied from additionalClaims.
  private static final Set<String> JWT_RESERVED_CLAIMS =
      Set.of(
          Claims.ISSUER,
          Claims.SUBJECT,
          Claims.AUDIENCE,
          Claims.EXPIRATION,
          Claims.NOT_BEFORE,
          Claims.ISSUED_AT,
          Claims.ID);
```

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/JwtTokenService.java
```diff
@@ -82,6 +83,17 @@ public class JwtTokenService implements Clock {
...
+
   public static List<String> getDefaultClaims() {
```
jnsfZ, line 97: Optional, this would be a Bad Pratice in Downstream Projects

**The problem:** nothing protects Essencium's own claim names. `createToken` writes `uid`, `roles`, `rights`, `locale` and `parent_token_id` onto the builder first and copies `additionalClaims` on top afterwards, so a user whose `getAdditionalClaims()` returns a key named `rights` replaces the real rights in the signed token. The authentication filter reads `rights` back out of that token to decide what a caller may do, so this grants or removes permissions. The new `JWT_RESERVED_CLAIMS` filter covers only the seven JWT standard names and does not help here.

**How exposed this is today:** `ApiTokenService` filters this list itself before calling `createToken`, so API tokens are fine. Access tokens from a normal login are not, because `toEssenciumUserDetails()` passes the entity's claims through unchanged. Nothing is broken right now, since no entity in this repository uses one of these names; it takes a downstream project choosing `rights`, `roles` or `uid` for a custom claim.

**What I would do:** filter this list in `createToken` as well, next to `JWT_RESERVED_CLAIMS`, so both paths are covered in the single method that builds tokens. The filter in `ApiTokenService` then becomes redundant and can go.

**What you need to decide:** first, whether a claim from `getAdditionalClaims()` may override `rights` at all — I would say no, but either way it should be a deliberate contract rather than an accident. Second, whether this belongs in this PR or a follow-up ticket; it touches two classes, so a follow-up is entirely reasonable.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/JwtTokenService.java
```diff
@@ -214,7 +226,10 @@ public String createToken(
                   .map(SessionToken::getId)
                   .orElse(null));
 
-      for (Map.Entry<String, Object> entry : userDetails.getAdditionalClaims().entrySet()) {
+      for (Map.Entry<String, Object> entry :
+          userDetails.getAdditionalClaims().entrySet().stream()
+              .filter(e -> !JWT_RESERVED_CLAIMS.contains(e.getKey()))
+              .toList()) {
```
jnsfZ: The stream builds a temporary list only to walk over it once. The original loop with a `continue` guard is the smaller change and behaves the same:

```suggestion
      for (Map.Entry<String, Object> entry : userDetails.getAdditionalClaims().entrySet()) {
        if (JWT_RESERVED_CLAIMS.contains(entry.getKey())) {
          continue;
        }
```

Dropping a claim is also silent, so a `log.debug` in that guard would help. A warning would be too noisy, since four are dropped on every API token creation.

jnsfZ (dismissed):

jnsfZ:
## Runtime test

Setup: `essencium-backend-development`, profiles `development,h2`, in-memory H2, access token lifetime 15 min. One script, run against a build of `2d601379` and of current `main`.

| Check | `main` | this branch |
|---|---|---|
| Login, `GET /v1/users/me`, `POST /auth/renew` | 200 | 200 |
| Access token claims (`uid`, `roles`, 20 `rights`) | present | present |
| API token as bearer, read | 200 | 200 |
| **API token `exp`, `validUntil` = +30d** | **15 min, inherited** | **29d 8h, correct** |
| API token `sub` | user e-mail | `<user>-api-token-<uuid>` |
| Create an API token while authenticated with one | 200 | 403, row persisted |
| `GET /v1/api-tokens/all` afterwards | 200 | **403 for every admin** |
| **`validUntil` = today, `exp` vs. now** | **+901s, usable** | **−55883s, 401** |

Row 4 is the fix: an API token expires at its configured `validUntil` instead of inheriting the creating user's access token.

### Finding 1 — a rejected API token creation persists a row that disables `/v1/api-tokens/all`

- **Trigger:** authenticate with an API token carrying `API_TOKEN`, call `POST /v1/api-tokens`.
- **Observed:** `403 UsernameNotFoundException: user '<user>-api-token-<uuid>' not found`, and the row is persisted.
- **Cause:** `ApiTokenService` sets `linkedUser` from the principal name, which is the token's synthetic name. `ApiTokenAssembler` resolves `linkedUser` through `loadUserByUsername`, which ends in `orElseThrow`.
- **Consequence:** `GET /v1/api-tokens/all` returns `403` for every admin while that row exists. `findAllAdmin` maps every row through the assembler, so one unresolvable `linkedUser` fails the entire response rather than that one entry. The status is `403` rather than `500` because `UsernameNotFoundException` is an `AuthenticationException`. `GET /v1/api-tokens` stays functional: its ownership filter compares `linkedUser` to the caller's e-mail, so the row is excluded (observed: 200, four tokens listed).
- **Recovery** (read from the code, not tested): `DELETE /v1/api-tokens/{id}` assembles no representation and should succeed, but the id is not obtainable through the API — `/all` fails, `GET /v1/api-tokens` filters the row out, and `GET /v1/api-tokens/{id}` assembles a representation. That leaves the database.
- **Permitted by configuration:** `ACCESS_TOKEN_ONLY_URLS` covers `/v1/users/me` and `/v1/users/me/**` only, so API tokens may call this endpoint. On `main` the call returns `200` and links the new token to the human user.
- **Open question, predating this PR:** whether an API token may create further tokens.
- **Defect independent of that question:** a rejected request leaves a persisted row behind, and that row makes an unrelated admin endpoint unusable with no API-level way to remove it.
- **Fix:** take `linkedUser` from the human user, or reject when the caller is an API token; `ApiTokenAssembler` should tolerate an unresolvable `linkedUser`.
- **Precondition:** `API_TOKEN` granted to a token.

### Finding 2 — `validUntil` = today yields a token expired on arrival

- **Trigger:** normal creation flow with an access token, `validUntil` set to today.
- **Observed:** `exp` 15 hours in the past, first request `401`.
- **Cause:** `validUntil` maps to `atStartOfDay()`, `ApiToken.isAccountNonExpired()` treats the whole day as valid, and `convertDtoToEntity` rejects only dates before today. The previous `exp` source masked this.
- **Fix:** `validUntil.plusDays(1).atStartOfDay()`.

jnsfZ (approved):

PR #971, opened 2026-07-31
fix: prevent PATCH of id, updatedBy and updatedAt

### LINKED ISSUE #869: fix: PATCH allows overwriting id, updatedBy and updatedAt
## Problem

The PATCH endpoint for partial entity updates only strips `createdBy` and `createdAt` from the update map before applying fields via reflection. The fields `id`, `updatedBy` and `updatedAt` are **not** protected and can be overwritten by any API client via PATCH.

This affects **all entities** extending `AbstractBaseModel` -- not just User, but every entity in the system.

## Context

- **File:** [`AbstractEntityService.java`](essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/AbstractEntityService.java), lines 210-219

```java
// only createdBy and createdAt are protected -- id, updatedBy, updatedAt are not
fieldUpdates.remove("createdBy");
fieldUpdates.remove("createdAt");

fieldUpdates.forEach((key, value) -> updateField(toUpdate, key, value));
```
- **Impact:**
  - Changing `id` can corrupt database relationships
  - Changing `updatedBy` can hide who actually made a modification
  - Changing `updatedAt` can backdate or future-date changes
- `AbstractUserService.patchPreProcessing` (line 228ff) also does **not** add any additional exclusions for these fields
- `updatedBy` and `updatedAt` are normally set automatically by Spring Data Auditing -- overwriting them via PATCH defeats that mechanism

## Implementation Hint

Consider making the protected field list a `protected` method instead of hardcoded `remove()` calls, so subclasses can extend it without duplicating the base list.

## Acceptance Criteria

- [x] The fields `id`, `updatedBy` and `updatedAt` are ignored in PATCH requests
- [x] Ideally the list of protected fields is extensible for subclasses (so `AbstractUserService` can protect additional fields like `password`, `loginDisabled`)
- [x] Test verifying that protected fields cannot be changed via PATCH
- [x] `CHANGELOG.md` updated
- [x] `MIGRATION.md` updated (if the fix changes behavior for downstream projects)

lwydra:
This PR adjusts `AbstractEntityService.java#patchPreProcessing` to ignore the meta-fields `id`, `updatedBy` and `updatedAt`.  
The resulting Set of ignored field names can be adjusted by downstream projects by overwritting `getPatchProtectedFieldNames`.

Closes #869.


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/AbstractEntityService.java
```diff
@@ -212,13 +213,24 @@ protected OUT patchPreProcessing(
...
+  /**
+   * Names of fields that must never be overwritten via PATCH, e.g. because they are managed
+   * automatically (audit fields) or identify the entity. Subclasses can extend this set to protect
+   * additional fields.
+   *
+   * @return the field names to strip from patch requests before applying them
+   */
+  @NotNull
+  protected Set<String> getPatchProtectedFieldNames() {
+    return Set.of("id", "createdBy", "createdAt", "updatedBy", "updatedAt");
```
pwannenmacher: Instead of hard-coded strings, it would be nice if we could use the JPA metamodels here (e.g. `AbstractBaseModel_.CREATED_BY`).

pwannenmacher (changes requested):

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/AbstractEntityService.java
pwannenmacher: We should also store the set in a static variable so that a new set does not have to be created every time it is called. We still need the overridable method to make the set available.

lwydra: Good suggestions, both done!

Regarding `id`: This field is implement by either `UUIDModel` or `SequenceIdModel` (depending on the chosen id-type) with both using the same variable name. As `Set.of` throws an exception when the provided attributes contain duplicates, I only used `UUIDModel_` here and added a comment to hint on the relationship

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/AbstractEntityService.java
```diff
@@ -212,13 +224,24 @@ protected OUT patchPreProcessing(
...
+   * Names of fields that must never be overwritten via PATCH, e.g. because they are managed
+   * automatically (audit fields) or identify the entity. Subclasses can extend this set to protect
+   * additional fields.
+   *
+   * @return the field names to strip from patch requests before applying them
+   */
+  @NotNull
+  protected Set<String> getPatchProtectedFieldNames() {
+    return PATCH_PROTECTED_FIELDS;
+  }
```
pwannenmacher, line 248: The Javadoc does not mention that the returned set is immutable. Please feel free to add `@return an unmodifiable set; copy it before extending` (or `@apiNote`).

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/AbstractEntityService.java
```diff
@@ -122,6 +125,15 @@ public abstract class AbstractEntityService<
               }),
           Map.entry(URI.class, URI::create));
 
+  private static final Set<String> PATCH_PROTECTED_FIELDS =
+      Set.of(
+          // Same value as SequenceIdModel_.ID
+          UUIDModel_.ID,
```
pwannenmacher: I’m afraid this is a rather fragile solution. The cleanest solution would probably be to add `String ID_FIELD = ‘id’;` to `Identifiable` and reference it — that’s the one type that all three share.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/service/AbstractEntityService.java
```diff
@@ -212,13 +224,24 @@ protected OUT patchPreProcessing(
     OUT out = repository.findById(id).orElseThrow(ResourceNotFoundException::new);
     final var toUpdate = (OUT) out.clone();
 
-    fieldUpdates.remove("createdBy");
-    fieldUpdates.remove("createdAt");
+    getPatchProtectedFieldNames().forEach(fieldUpdates::remove);
 
     fieldUpdates.forEach((key, value) -> updateField(toUpdate, key, value));
```
pwannenmacher: To avoid triggering an `UnsupportedOperationException` when working with an immutable map, we should use the following version:
```java
  final var sanitized = new HashMap<>(fieldUpdates);
  sanitized.keySet().removeAll(getPatchProtectedFieldNames());
  sanitized.forEach((key, value) -> updateField(toUpdate, key, value));
```

pwannenmacher (changes requested):
In AbstractUserController.java:358-361 we have a second Version of Sanitation. Maybe we should migrate it to the new mechanism, too.

lwydra:
> In AbstractUserController.java:358-361 we have a second Version of Sanitation. Maybe we should migrate it to the new mechanism, too.

Done - moving this into the service now also produces a breaking change for `DefaultUserInitializer` though, as this one uses `userService.patch` as well (l. 171).

pwannenmacher (approved):

PR #990, opened 2026-08-18
fix: keep framework status codes in ProblemDetail error handling

pwannenmacher:
Follow-up to #929.

## Problem

`GlobalExceptionHandler`'s `@ExceptionHandler(Exception.class)` catch-all matches every exception without a more specific handler, and `ExceptionHandlerExceptionResolver` runs before `DefaultHandlerExceptionResolver`. Verified against the merged state:

| Exception | expected | before this PR |
| --- | --- | --- |
| `AccessDeniedException` from `@Secured` | 403 | 500 |
| `MethodArgumentTypeMismatchException` | 400 | 500 |
| `HttpRequestMethodNotSupportedException` | 405 | 500 |

Same for 415, 404 and the missing-request-value exceptions. The replaced `RestExceptionHandler` extended `ResponseEntityExceptionHandler` and covered these; the new one did not. Authorization failures were indistinguishable from server errors, and each one logged an `ERROR` with a stack trace.

Separately, the `ResponseStatusException` and `MissingRequestCookieException` handlers overwrote `detail` after `ProblemDetailFactory` had sanitised it. With the default `include-message=never`, a reason like `ResponseStatusException(NOT_FOUND, "internal reason 4711")` was returned verbatim, and a `ResponseStatusException` without a reason lost the `"An error occurred"` fallback entirely.

## Changes

- `GlobalExceptionHandler` extends `ResponseEntityExceptionHandler`. The framework keeps its status code and headers; only the body is rewritten to a `ProblemDetail` in `handleExceptionInternal`.
- `AccessDeniedException` gets an explicit handler: 403 for an authenticated caller, 401 for an anonymous one, mirroring `ExceptionTranslationFilter`.
- The `ResponseStatusException` and `MissingRequestCookieException` handlers are gone; both paths run through the parent handler and the factory like every other error. `ServletRequestBindingException` is covered as a whole instead of only its cookie subclass.
- `createResponse` and `ProblemDetailFactory.create` take a `HttpStatusCode`, so a non-standard status no longer throws inside the handler.
- `essencium.error.urn-prefix` is validated at startup instead of failing inside the exception handler.
- New `ErrorCode`s `METHOD_NOT_ALLOWED`, `NOT_ACCEPTABLE`, `PAYLOAD_TOO_LARGE` and `UNSUPPORTED_MEDIA_TYPE` replace the generic `RESPONSE_STATUS_EXCEPTION`, which is removed.
- `MIGRATION.md` and `CHANGELOG.md` document the `ProblemDetail` migration, which had no entry yet.

## Tests

- `GlobalExceptionHandlerWebTest` drives the real MVC resolver chain; the mock-based test calls handler methods directly and cannot tell whether an exception reaches its handler at all.
- `RoleControllerIntegrationTest` covers a `@Secured` denial end to end. The existing 403 tests all raise `NotAllowedException`, so none of them touched this path.
- `AccessEntityFilteringIntegrationTest` still asserted the removed `$.message` shape (added in #985, not adjusted by #929) and **was failing on `main`**.

738 unit tests and 190 integration tests pass.

## Note on the integration test config

`server.error.include-message` in `src/testIntegration/resources/application.yaml` had no effect — Spring Boot 4 moved `ErrorProperties` from `ServerProperties` to `WebProperties`, and `ServerProperties` has no `error` property at all. The live key was already set further down the same file, so the block was a stale duplicate suggesting a different configuration than the tests actually run with.


essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
+      return ErrorCode.INTERNAL_SERVER_ERROR;
+    }
+
+    return switch (statusCode.value()) {
+      case 404 -> ErrorCode.NOT_FOUND;
+      case 405 -> ErrorCode.METHOD_NOT_ALLOWED;
+      case 406 -> ErrorCode.NOT_ACCEPTABLE;
+      case 413 -> ErrorCode.PAYLOAD_TOO_LARGE;
+      case 415 -> ErrorCode.UNSUPPORTED_MEDIA_TYPE;
+      default -> ErrorCode.INVALID_INPUT;
```
jnsfZ: **401, 403 and 409 resolve to `INVALID_INPUT`**

The switch enumerates 404, 405, 406, 413 and 415; every other status reaches the default arm. That includes 401, 403 and 409, for which `AUTHENTICATION_FAILED`, `FORBIDDEN` and `DUPLICATE_RESOURCE` exist in `ErrorCode`.

Observed for `ResponseStatusException`, which reaches this switch through the parent handler:

```
401 -> type urn:frachtwerk:error:INVALID_INPUT
403 -> type urn:frachtwerk:error:INVALID_INPUT
409 -> type urn:frachtwerk:error:INVALID_INPUT
```

`AuthenticationController` throws `ResponseStatusException` with `UNAUTHORIZED` at lines 140, 158, 179 and 182, and with `FORBIDDEN` at line 170, so `/auth/token` and `/auth/renew` are on this path.

`type` is the member RFC 9457 designates as the client's dispatch key, and MIGRATION.md directs clients to read it. A caller branching on `AUTHENTICATION_FAILED` does not match a 401 arriving here, and the same logical condition reports different `type` values depending on which handler produced the response — a `NotAllowedException` and a `ResponseStatusException(FORBIDDEN)` both yield 403 but different codes.

The default arm also covers 3xx and any non-standard status, which are labelled as client input errors.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
+    if (exception instanceof HttpMessageNotReadableException) {
+      return ErrorCode.MALFORMED_REQUEST;
+    }
 
-    return validationResult.getResolvableErrors().stream()
-        .map(MessageSourceResolvable::getDefaultMessage)
-        .map(message -> new FieldErrorResponse(field, message));
+    if (exception instanceof BindException
+        || exception instanceof HandlerMethodValidationException
+        || exception instanceof MethodValidationException) {
```
jnsfZ: **`MethodValidationException` is dispatched with a 500 status**

These type checks precede the `is5xxServerError()` guard four lines below. `ResponseEntityExceptionHandler.handleException` dispatches `MethodValidationException` with `HttpStatus.INTERNAL_SERVER_ERROR` (spring-webmvc 7.0.8, line 198), so it matches this branch while carrying a 500.

The resulting body is `status: 500` with `type: urn:frachtwerk:error:VALIDATION_FAILED`.

`MethodValidationException` reports a constraint violation on a server-side method invocation — a wiring or programming fault rather than malformed input. A body pairing a 500 with a validation code asks the caller to correct a payload it cannot influence, and presents a server defect under a client-error code.

The javadoc above the method describes the type-before-status ordering as keeping "the 400s" distinguishable, which does not cover this case.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
+  @Override
+  protected ResponseEntity<Object> handleExceptionInternal(
+      Exception exception,
+      @Nullable Object body,
+      HttpHeaders headers,
+      HttpStatusCode statusCode,
+      WebRequest request) {
+    HttpServletRequest servletRequest = servletRequest(request);
+
+    if (statusCode.is5xxServerError()) {
```
jnsfZ: **4xx responses routed through this method produce no log record**

Logging is gated on `statusCode.is5xxServerError()`, and `validationResponse` logs at no status. The exception types the parent routes here — both media-type exceptions, `TypeMismatchException`, the missing-parameter/part/path-variable family, `ServletRequestBindingException`, `ErrorResponseException` and `ResponseStatusException` — therefore leave no trace when they resolve to a 4xx.

Observed across the 15 cases of `GlobalExceptionHandlerWebTest`: the only output is two `ERROR Unhandled exception` lines from the two 500 cases, plus one `WARN` from Spring's own `PageNotFound` logger for the 405.

With `spring.web.error.include-message` at its default of `never`, the corresponding response is:

```json
{"type":"...INVALID_INPUT","title":"Bad Request","status":400,
 "detail":"An error occurred","instance":"/v1/users","timestamp":"..."}
```

Neither the response nor the log names the exception type or the offending field, and no log level exposes it.

Some of the affected types indicate a server-side fault rather than a client mistake — `HttpMediaTypeNotAcceptableException` from a content-negotiation misconfiguration, `TypeMismatchException` from a faulty `Converter`.

A related path: `AuthenticationController:138` catches `AuthenticationException` and rethrows `ResponseStatusException(UNAUTHORIZED, ...)`. `DaoAuthenticationProvider.retrieveUser` (spring-security-core 7.1.0, lines 104-121) wraps any exception from `loadUserByUsername` — connection failures, an unreachable LDAP server, an NPE in a user service — into `InternalAuthenticationServiceException`, a subtype of `AuthenticationException`. `ProviderManager` logs it at DEBUG, and `AuthenticationFailureListener:36` subscribes only to `AuthenticationFailureBadCredentialsEvent`. An authentication-backend outage therefore appears as an unlogged 401 with `detail: "An error occurred"`, indistinguishable from a wrong password.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
+    ProblemDetail problemDetail =
+        problemDetailFactory.create(
+            statusCode,
+            resolveErrorCode(exception, statusCode),
+            detailOf(body, exception),
+            exception,
+            servletRequest);
 
-    return response;
+    return ResponseEntity.status(statusCode).headers(headers).body(problemDetail);
```
jnsfZ: **Three behaviours of the parent `handleExceptionInternal` are not reproduced**

`ResponseEntityExceptionHandler.handleExceptionInternal` (spring-webmvc 7.0.8, lines 617-638) additionally:

- returns `null`, with a warning, when `response.isCommitted()`
- calls `errorResponse.updateAndGetBody(messageSource, locale)` when `body` is `null`
- delegates to `createResponseEntity(body, headers, statusCode, request)`

Possible consequences:

1. `AsyncRequestTimeoutException` (503) is routed here by the parent (line 407). For a streaming or SSE response the response is usually already committed at that point, so returning a body attempts a second write. `AsyncRequestNotUsableException` is unaffected — the parent returns `null` at line 562 before reaching this method.
2. The class is `MessageSourceAware` through its parent and the project supplies a `MessageSource` (`ErrorMessageLocalization`). Without `updateAndGetBody`, `problemDetail.<ExceptionType>` message codes are not resolved for the parent handlers that pass `body == null`: `ErrorResponseException`, both media-type exceptions, the missing-parameter/part/path-variable family, `NoHandlerFoundException`, `NoResourceFoundException`, `AsyncRequestTimeoutException`, `MaxUploadSizeExceededException`.
3. `createResponseEntity` is named in the parent's class javadoc (line 68) as the hook for subclasses to inspect or modify body, headers and status. A subclass override of it has no effect on any exception routed through this method.

The request-attribute write above is also broader than the parent's, which is conditioned on `statusCode.equals(HttpStatus.INTERNAL_SERVER_ERROR) && body == null`. `AsyncRequestTimeoutException` consequently logs at ERROR with a stack trace and reports `type: INTERNAL_SERVER_ERROR` on a 503 body.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
+            status, ErrorCode.VALIDATION_FAILED, "Validation failed", exception, servletRequest);
 
-      current = current.getCause();
+    problemDetailFactory.addFieldErrorsIfAllowed(problemDetail, fieldErrors, servletRequest);
+
+    return ResponseEntity.status(status).headers(headers).body(problemDetail);
+  }
+
+  /** {@code getMessage()} carries the status code as a prefix, the problem detail does not. */
+  private static @Nullable String detailOf(@Nullable Object body, Exception exception) {
```
jnsfZ: **`getMessage()` fallback exposes framework-internal strings**

The `ErrorResponse` branch requires `getBody().getDetail() != null`. `ResponseStatusException` derives its detail from `reason`, so a construction without a reason leaves it `null` and execution falls through to `exception.getMessage()`.

`ResponseStatusException.getMessage()` returns the status plus the quoted reason; `ErrorResponseException.getMessage()` returns the status, the headers and `ProblemDetail.toString()`. Observed with `include-message: always`:

```
ResponseStatusException(NOT_FOUND)                  -> detail: "404 NOT_FOUND"
ErrorResponseException(HttpStatusCode.valueOf(499)) -> detail: "499, ProblemDetail[type='null', title='null', status=499, detail='null', instance='null', properties='null']"
```

`throw new ResponseStatusException(HttpStatus.NOT_FOUND)` without a reason is a common form, so the first case is reachable from library and application code alike. The second places an internal object rendering in a client-facing field.

The method's own javadoc identifies `getMessage()` as carrying the status prefix, which is the behaviour the fallback then depends on.

The 499 response also carries no `title`: `ProblemDetail.getTitle()` derives it from `HttpStatus.resolve(status)`, which is `null` outside the enum, while `HttpStatusCode.valueOf` accepts 100 through 999. Under the default `include-message: never` such a response has neither a `title` nor a usable `detail`.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
+            resolveErrorCode(exception, statusCode),
+            detailOf(body, exception),
+            exception,
+            servletRequest);
 
-    return response;
+    return ResponseEntity.status(statusCode).headers(headers).body(problemDetail);
   }
 
   protected ResponseEntity<ProblemDetail> createResponse(
```
jnsfZ: **`createResponse` is not on the path taken by the inherited handlers**

MIGRATION.md presents this method as the extension point: "subclass `GlobalExceptionHandler`, implement `ProblemErrorCode` for your own codes and call the `protected createResponse(...)`".

`handleExceptionInternal` (line 274) and `validationResponse` call `problemDetailFactory.create(...)` directly rather than through this method. An override therefore takes effect for the eleven library-specific `@ExceptionHandler` methods and for none of the roughly twenty types inherited from `ResponseEntityExceptionHandler`, nor for validation errors. A consumer adding a correlation id or a tenant-specific `type` here would see it applied to part of the error surface, with nothing indicating the split.

The parameter list also has no `HttpHeaders`, so responses that require a header cannot be produced through this seam — `WWW-Authenticate` on a 401 (RFC 9110 §15.5.2), `Allow` on a 405, `Retry-After` on a 429. The inherited path does thread `headers` through, so the two paths differ in capability as well as in identity.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -238,90 +212,167 @@ public ResponseEntity<ProblemDetail> handleAuthenticationException(
...
-    AuthenticationException authenticationException =
-        findCause(exception, AuthenticationException.class);
-
-    if (authenticationException != null) {
-      return handleAuthenticationException(authenticationException, request);
+  /** Mirrors {@code ExceptionTranslationFilter}, which no longer sees the exception. */
+  @ExceptionHandler(AccessDeniedException.class)
+  public ResponseEntity<ProblemDetail> handleAccessDeniedException(
+      AccessDeniedException exception, HttpServletRequest request) {
+    if (isAuthenticated(SecurityContextHolder.getContext().getAuthentication())) {
```
jnsfZ: **Divergence from `ExceptionTranslationFilter`, which the javadoc names as the model**

`ExceptionTranslationFilter.handleAccessDeniedException` (spring-security-web 7.1.0, lines 188-212) differs in five respects:

- it branches on the injectable `AuthenticationTrustResolver`, which classifies **remember-me** authentications as untrusted alongside anonymous ones. `isAuthenticated()` (line 361) excludes only `AnonymousAuthenticationToken`, so a remember-me principal is treated as authenticated here and as unauthenticated there.
- its anonymous branch calls `sendStartAuthentication`, which clears the `SecurityContext`, stores the request in the `RequestCache` and invokes the configured `AuthenticationEntryPoint` (lines 218-221). None of that occurs here, so a redirect back to the original request after login is unavailable under the `oauth` profile.
- its authenticated branch delegates to the configured `AccessDeniedHandler` (line 210). An application registering one through `exceptionHandling(...)` does not see it invoked for a denial raised inside the dispatch.
- it reads the context through the pluggable `SecurityContextHolderStrategy`; this line uses the static `SecurityContextHolder`.
- the 401 produced here carries no `WWW-Authenticate` challenge, which the filter path obtains from the entry point.

Neither branch records the denial. `ExceptionTranslationFilter` logs at TRACE, and this handler resolves the exception before the filter observes it, so a `@Secured` or `@PreAuthorize` denial leaves no trace — for a library providing RBAC, that is the signal separating a misconfigured client from privilege probing.

On reachability: `DEFAULT_PROTECTED_URLS` requires `.authenticated()` at filter level and `JwtTokenAuthenticationFilter` short-circuits ahead of it, so an anonymous request does not arrive here. Observed without a token on `/v1/roles` and `/v1/api-tokens`: `401`, empty body, no `Content-Type`. The 401 branch is reachable only where a URL is permitted at filter level and guarded by method security.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
```diff
@@ -30,25 +30,50 @@
...
+/**
+ * Maps exceptions to RFC 9457 {@link ProblemDetail} responses. Extends {@link
+ * ResponseEntityExceptionHandler} so the Spring MVC exceptions keep their status code.
+ *
+ * <p>Deliberately has no {@code @ExceptionHandler(Exception.class)}: such a mapping would match
+ * every exception, and {@code ExceptionHandlerMethodResolver} only walks {@code getCause()} when
+ * the thrown type itself has no mapping. A wrapped exception would therefore never reach its
+ * specific handler. The catch-all lives in {@link FallbackExceptionHandler}, which runs last.
+ */
+@Order(Ordered.LOWEST_PRECEDENCE - 1)
```
jnsfZ: **Precedence relative to application advices**

`LOWEST_PRECEDENCE - 1` here and `LOWEST_PRECEDENCE` on `FallbackExceptionHandler` order the two library advices against each other. An application `@ControllerAdvice` that declares no `@Order` also sits at `LOWEST_PRECEDENCE` and therefore ranks below this class.

`ExceptionHandlerExceptionResolver.getExceptionHandlerMethod` returns the first advice holding a mapping for the thrown type. Through the parent, this class claims these types in addition to its own `@ExceptionHandler` methods:

`HttpRequestMethodNotSupportedException`, `HttpMediaTypeNotSupportedException`, `HttpMediaTypeNotAcceptableException`, `MissingPathVariableException`, `MissingServletRequestParameterException`, `MissingServletRequestPartException`, `ServletRequestBindingException`, `NoHandlerFoundException`, `NoResourceFoundException`, `AsyncRequestTimeoutException`, `ErrorResponseException`, `MaxUploadSizeExceededException`, `ConversionNotSupportedException`, `TypeMismatchException`, `HttpMessageNotWritableException`, `MethodValidationException`, `AsyncRequestNotUsableException`.

An application advice handling any of those without an explicit `@Order` loses the mapping to this class, and the response is produced here instead. The ordering requirement an application needs to observe is not stated in MIGRATION.md.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/FallbackExceptionHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+  private static final Logger log = LoggerFactory.getLogger(FallbackExceptionHandler.class);
+
+  private final ProblemDetailFactory problemDetailFactory;
+
+  public FallbackExceptionHandler(ProblemDetailFactory problemDetailFactory) {
+    this.problemDetailFactory = problemDetailFactory;
+  }
+
+  @ExceptionHandler(Exception.class)
```
jnsfZ: **`@ResponseStatus` on application exception classes is not consulted**

`@ExceptionHandler(Exception.class)` matches any exception lacking a more specific mapping, including exception classes annotated `@ResponseStatus`. `ExceptionHandlerExceptionResolver` precedes `ResponseStatusExceptionResolver` in the resolver chain, so the annotation is never read.

Observed with a `@ResponseStatus(HttpStatus.CONFLICT)` exception thrown from a controller:

```
with the library advices registered -> 500, type urn:frachtwerk:error:INTERNAL_SERVER_ERROR
without them                        -> 409
```

`@ResponseStatus` on an exception class is a standard Spring mechanism and is in use in consuming projects; two present examples are an exception annotated `@ResponseStatus(CONFLICT)` and one annotated `@ResponseStatus(BAD_REQUEST)` extending `jakarta.validation.ValidationException`. Both resolve to 500, and the accompanying `ERROR Unhandled exception` describes an application-declared status as unhandled.

This is the same shape as the condition the pull request addresses for the Spring MVC exceptions — a declared status flattened to 500 — applied to statuses declared by application code. MIGRATION.md covers the removal of `@ResponseStatus` from the library's own exceptions, but not the effect on application exception classes carrying it.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
```diff
@@ -51,9 +52,9 @@ public ProblemDetailFactory(
...
 
   public ProblemDetail create(
-      HttpStatus status,
+      HttpStatusCode status,
       ProblemErrorCode errorCode,
-      String detail,
+      @Nullable String detail,
       Throwable throwable,
       HttpServletRequest request) {
     String resolvedDetail = resolveDetail(detail, request);
```
jnsfZ: **Neither URI is validated before construction**

Line 64 builds `type` by concatenating `essencium.error.urn-prefix` with `errorCode.getCode()`; line 65 builds `instance` from `request.getRequestURI()`. Neither result is checked.

`EssenciumErrorProperties.isUrnPrefixParsable` validates the prefix against a fixed `PROBE` suffix at startup, which places no constraint on the code half. `ProblemErrorCode` is a public interface with a single `String getCode()` and no documented character set, and MIGRATION.md invites applications to implement it. A code containing a space or an unescaped `%` produces:

```
IllegalArgumentException: Illegal character in opaque part at index 23: urn:frachtwerk:error:MY CODE
```

thrown from inside an `@ExceptionHandler` invocation. `ExceptionHandlerExceptionResolver.doResolveHandlerMethodException` (lines 461-472) catches that, logs at WARN on a Spring logger category and returns `null`, continuing with default processing of the original exception — so the caller receives the container error response rather than `problem+json`. Because every handler except the fallback builds the body before logging, the original exception is not recorded either.

A `getCode()` returning `null` or an empty string is accepted and serialised as `urn:frachtwerk:error:null` and as the bare prefix respectively.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
```diff
@@ -73,7 +74,7 @@ public ProblemDetail create(
 
   public void addFieldErrorsIfAllowed(
       ProblemDetail problemDetail,
-      List<FieldErrorResponse> fieldErrors,
+      @Nullable List<FieldErrorResponse> fieldErrors,
       HttpServletRequest request) {
     if (shouldIncludeMessage(request) && fieldErrors != null && !fieldErrors.isEmpty()) {
```
jnsfZ: **`fieldErrors` is gated on `include-message` rather than `include-binding-errors`**

The condition reads `shouldIncludeMessage`, i.e. `spring.web.error.include-message`. `ErrorProperties` also exposes `include-binding-errors` (`getIncludeBindingErrors()`, line 97), the setting Spring Boot provides for exactly this decision, and nothing in the codebase reads it.

`src/testIntegration/resources/application.yaml:134` sets `spring.web.error.include-binding-errors: always`, so the integration-test configuration declares a value with no effect.

Two configurations then behave contrary to what the property names suggest: `include-binding-errors: never` together with `include-message: always` still emits field names, and `include-binding-errors: always` together with `include-message: never` suppresses them. An operator suppressing internal field names from responses while keeping human-readable detail cannot express that.

MIGRATION.md documents `fieldErrors` as following `include-message`, which matches the implementation but not the Spring Boot property covering the same concern.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/configuration/properties/EssenciumErrorProperties.java
```diff
@@ -29,5 +32,23 @@
 @ConfigurationProperties(prefix = "essencium.error")
 public class EssenciumErrorProperties {
 
+  private static final String PROBE_CODE = "PROBE";
+
   @NotBlank private String urnPrefix = "urn:frachtwerk:error:";
+
+  /** Fails startup rather than letting {@code URI.create} throw inside the exception handler. */
+  @JsonIgnore
+  @AssertTrue(message = "must form a valid URI together with an error code")
```
jnsfZ: **Constraint reporting and the chosen predicate**

Two observations about expressing this rule as a derived boolean getter.

*The reported property is the getter, not the configured key.* `BindValidationFailureAnalyzer` renders `Property: <objectName>.<field>`, and Bean Validation derives the field name from the getter, so a malformed prefix surfaces as:

```
Property: essencium.error.urnPrefixParsable
Value: "false"
Reason: must form a valid URI together with an error code
```

The key to change (`essencium.error.urn-prefix`) and the value that failed are both absent. The `catch` at line 50 additionally discards the `IllegalArgumentException` message, which names the offending character and its index.

*The predicate is parsability, not absoluteness.* RFC 9457 states that `type` should be an absolute URI. These prefixes satisfy the current check while yielding a relative reference:

```
"x"       -> xPROBE       (scheme null)
"//"      -> //PROBE      (scheme null)
"#frag:"  -> #frag:PROBE  (scheme null)
```

A conforming client resolves a relative `type` against the request URI, so the same error reports a different `type` per endpoint.

`urnPrefix` also retains a public setter, and `ProblemDetailFactory` reads `getUrnPrefix()` on every error rather than once, so the state validated at startup is not the state necessarily in effect later.

MIGRATION.md
```diff
@@ -2,6 +2,18 @@
 
 ## Version `4.0.0-SNAPSHOT` (unreleased)
 
+### RFC 9457 `ProblemDetail` error responses
+
+All error responses are now `application/problem+json` (`GlobalExceptionHandler` + `ProblemDetailFactory`); the `ErrorResponse` DTO and `RestExceptionHandler` are gone.
```
jnsfZ: **The media-type statement does not hold for responses written outside the dispatch**

Bodies produced by the two advices do carry `application/problem+json`. Responses written before or outside handler dispatch do not pass through them.

`JwtTokenAuthenticationFilter.unsuccessfulAuthentication` delegates to `JwtAuthenticationFailureHandler`, which calls `response.sendError(401, message)`. `sendError` triggers an ERROR dispatch to `/error`, served by Boot's `BasicErrorController`, whose body is `{timestamp, status, error, path}`. The codebase contains no custom `ErrorController` or `ErrorAttributes`, and `spring.mvc.problemdetails.enabled` is not set anywhere.

Observed without a token:

```
GET /v1/roles      -> 401, no Content-Type, empty body
GET /v1/api-tokens -> 401, no Content-Type, empty body
```

An absent or expired token is the most frequent client-facing error, and these responses carry no `type`, `instance` or `detail`. A client following the adjacent instruction to read `detail` instead of `message` finds neither field here.

Filter-level authorization denials reach the response through `ExceptionTranslationFilter` and its entry point or `AccessDeniedHandler`, likewise outside the advices.

essencium-backend/src/test/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandlerWebTest.java
```diff
@@ -0,0 +1,292 @@
...
+      mockMvc.perform(get("/probe/param")).andExpect(status().isBadRequest());
+    }
+
+    @Test
+    void missingCookieIsBadRequest() throws Exception {
+      mockMvc.perform(get("/probe/cookie")).andExpect(status().isBadRequest());
+    }
+
+    @Test
+    void responseStatusExceptionKeepsItsStatus() throws Exception {
```
jnsfZ: **Two coverage gaps relative to the contract being established**

*The `type` for statuses outside the switch is unasserted.* This case uses `HttpStatus.NOT_FOUND`, one of the five statuses `resolveErrorCode` enumerates. The statuses reaching the default arm — 401 and 403 among them, both raised by `AuthenticationController` — are not exercised here.

*The response media type is asserted nowhere in the repository.* `problem+json` and `APPLICATION_PROBLEM_JSON` do not appear in any test. The value depends on content negotiation: `AbstractMessageConverterMethodProcessor` applies `application/problem+json` only when no requested type is compatible, so a request sending `Accept: application/json` receives `application/json` while a request without `Accept` receives `application/problem+json`. MIGRATION.md presents the media type as part of the contract.

Also worth noting for scope: `standaloneSetup` builds no security filter chain, so this class cannot distinguish an exception resolved by an advice from one resolved inside the filter chain — the open question for `AccessDeniedException`. The two `denied` cases set `SecurityContextHolder` directly and throw from the controller, exercising the handler rather than the routing between handler and filter.

essencium-backend/src/testIntegration/java/de/frachtwerk/essencium/backend/test/integration/controller/RoleControllerIntegrationTest.java
```diff
@@ -136,6 +137,20 @@ public void tearDownSingle() {
     rightRepository.deleteAll(testRights);
   }
 
+  @Test
+  void testAccessDeniedByMethodSecurityIsForbidden() throws Exception {
+    Role roleWithoutRights =
+        roleRepository.save(Role.builder().name("TEST_ROLE_NO_RIGHTS").build());
```
jnsfZ: **Fixtures created here are not removed**

`TEST_ROLE_NO_RIGHTS` and `no.rights@frachtwerk.de` are created in this test; `tearDownSingle` removes `testEditableRole`, `testProtectedRole` and `testRights` only.

Both rows outlive the last test of the class and remain in the Testcontainers database, which is shared with the other integration-test classes. Role names and usernames are unique, so a second execution against the same database instance conflicts, and classes that enumerate roles or users observe the extra rows.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The switch covers 404, 405, 406, 413 and 415. Everything else falls to `INVALID_INPUT`, including 401, 403 and 409 — while `AUTHENTICATION_FAILED`, `FORBIDDEN` and `DUPLICATE_RESOURCE` exist in `ErrorCode`.

`AuthenticationController` throws `ResponseStatusException` with `UNAUTHORIZED` (lines 140, 158, 179, 182) and `FORBIDDEN` (line 170), so `/auth/token` and `/auth/renew` return:

```
401 -> type urn:frachtwerk:error:INVALID_INPUT
403 -> type urn:frachtwerk:error:INVALID_INPUT
```

Clients dispatch on `type`, so a 401 from these endpoints does not match `AUTHENTICATION_FAILED`. A 403 from `NotAllowedException` and a 403 from `ResponseStatusException` also report different codes for the same condition.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: These type checks run before the `is5xxServerError()` guard below. The parent dispatches `MethodValidationException` with a 500 status, so it matches here and the response becomes `status: 500` with `type: VALIDATION_FAILED`.

`MethodValidationException` means a constraint failed on a server-side method call, not that the request was wrong.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: Logging happens only for 5xx. The 4xx cases the parent routes here — media type, type mismatch, missing parameter/part/path variable, binding, `ResponseStatusException` — produce no log line.

With `include-message` at its default `never`, the response is `detail: "An error occurred"` with no exception type and no field name. Nothing then records what failed, in the response or in the log.

Some of those cases are server-side faults rather than client mistakes, e.g. `HttpMediaTypeNotAcceptableException` from a content-negotiation misconfiguration.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The parent's `handleExceptionInternal` also returns `null` when the response is already committed, resolves the detail through `MessageSource` when `body` is null, and delegates to `createResponseEntity`. This override does none of the three.

That means an `AsyncRequestTimeoutException` on a streaming response attempts a second write, `problemDetail.<ExceptionType>` message codes are not resolved for the handlers that pass a null body, and a subclass overriding `createResponseEntity` has no effect on this path.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The `ErrorResponse` branch requires a non-null detail. `ResponseStatusException` takes its detail from `reason`, so one constructed without a reason falls through to `exception.getMessage()`:

```
ResponseStatusException(NOT_FOUND) -> detail "404 NOT_FOUND"
ErrorResponseException(499)        -> detail "499, ProblemDetail[type='null', title='null', status=499, ...]"
```

`throw new ResponseStatusException(HttpStatus.NOT_FOUND)` is a common form. The second case puts an internal object dump into a client-facing field.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: MIGRATION.md points applications at this method as the extension point. `handleExceptionInternal` and `validationResponse` call `problemDetailFactory.create(...)` directly, so an override applies to the handlers declared in this class but not to the types inherited from `ResponseEntityExceptionHandler` or to validation errors.

There is also no `HttpHeaders` parameter, so a 401 with `WWW-Authenticate` or a 405 with `Allow` cannot be produced through it.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The javadoc says this mirrors `ExceptionTranslationFilter`. Two differences that affect applications:

- The filter delegates to the configured `AccessDeniedHandler`, and to the `AuthenticationEntryPoint` for anonymous callers. Neither runs here, so an application that registers one does not see it used for denials raised inside the dispatch.
- The filter treats remember-me as untrusted via `AuthenticationTrustResolver`; `isAuthenticated()` excludes only `AnonymousAuthenticationToken`.

Neither branch logs the denial, so a `@Secured` rejection leaves no trace.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/FallbackExceptionHandler.java
jnsfZ: `Exception.class` also matches exception classes annotated `@ResponseStatus`, and `ExceptionHandlerExceptionResolver` runs before `ResponseStatusExceptionResolver`, so the annotation is never read:

```
@ResponseStatus(CONFLICT) exception, advices registered -> 500
same exception, advices absent                         -> 409
```

Consuming projects use `@ResponseStatus` on exception classes. This is the case the PR fixes for the Spring MVC exceptions — a declared status becoming 500 — applied to statuses declared by application code.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
jnsfZ: `fieldErrors` is gated on `include-message`. Spring Boot's property for this is `include-binding-errors`, which nothing reads — `src/testIntegration/resources/application.yaml:134` sets it and it has no effect.

So `include-binding-errors: never` still emits field names, and suppressing field names while keeping a readable `detail` is not expressible.

MIGRATION.md
jnsfZ: Bodies produced by the advices are `problem+json`. Responses written outside handler dispatch are not.

`JwtAuthenticationFailureHandler` calls `response.sendError(401, ...)`, which dispatches to `/error` and Boot's `BasicErrorController`, whose body is `{timestamp, status, error, path}`. Measured without a token on `/v1/roles`: 401, no `Content-Type`, empty body.

A missing or expired token is the most common client-facing error, and it carries no `type`, `instance` or `detail`.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The switch covers 404, 405, 406, 413 and 415. Everything else reaches `INVALID_INPUT`, including 401, 403 and 409 — while `AUTHENTICATION_FAILED`, `FORBIDDEN` and `DUPLICATE_RESOURCE` exist in `ErrorCode`.

`AuthenticationController` throws `ResponseStatusException` with `UNAUTHORIZED` (lines 140, 158, 179, 182) and `FORBIDDEN` (line 170). A failed login against the running application returns:

```
POST /auth/token, wrong password
401 {"detail":"Bad credentials","status":401,"title":"Unauthorized",
     "type":"urn:frachtwerk:error:INVALID_INPUT"}
```

Clients dispatch on `type`, so this does not match `AUTHENTICATION_FAILED`. A 403 raised as `NotAllowedException` and a 403 raised as `ResponseStatusException` also report different codes for the same condition.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: Logging happens only for 5xx. The 4xx cases the parent routes here — media type, type mismatch, missing parameter/part/path variable, binding, `ResponseStatusException` — produce no log line.

The failed login above is one of them: a 401 from `/auth/token` is not logged at any level. With `include-message` at its default `never` the body is `detail: "An error occurred"`, so neither the response nor the log identifies what failed.

`AuthenticationController:138` also catches `AuthenticationException` broadly. `InternalAuthenticationServiceException` — raised when `loadUserByUsername` fails, e.g. an unreachable LDAP server or a database error — is a subtype, so an authentication backend outage takes the same silent 401 path.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The `ErrorResponse` branch requires a non-null detail. `ResponseStatusException` takes its detail from `reason`, so one constructed without a reason falls through to `exception.getMessage()`, which returns the status and the quoted reason:

```
ResponseStatusException(NOT_FOUND) -> detail "404 NOT_FOUND"
```

The framework's own exceptions supply a detail, so this concerns `ResponseStatusException` raised without a reason in application code — a common form. The method's javadoc notes that `getMessage()` carries the status prefix.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: MIGRATION.md points applications at this method as the extension point. `handleExceptionInternal` and `validationResponse` call `problemDetailFactory.create(...)` directly, so an override applies to the handlers declared in this class but not to the types inherited from `ResponseEntityExceptionHandler` or to validation errors.

There is also no `HttpHeaders` parameter, so a 401 with `WWW-Authenticate` or a 405 with `Allow` cannot be produced through it.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The parent's `handleExceptionInternal` returns `null` when the response is already committed, resolves the detail through `MessageSource` when `body` is null, and delegates to `createResponseEntity`. This override does none of the three.

The first two are not reachable in the library as it stands — there are no async or streaming endpoints, and no `problemDetail.*` message keys are defined. The third applies now: `createResponseEntity` is named in the parent's javadoc as the hook for adjusting body, headers or status, and a subclass overriding it has no effect on any exception routed through here.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: Neither branch logs the denial, and `ExceptionTranslationFilter` logs at TRACE, so a `@Secured` or `@PreAuthorize` rejection leaves no record.

The javadoc says this mirrors `ExceptionTranslationFilter`. The filter delegates to the configured `AccessDeniedHandler`, and to the `AuthenticationEntryPoint` for anonymous callers; neither runs here, so an application registering one does not see it used for denials raised inside the dispatch. No consumer configures one at present.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/FallbackExceptionHandler.java
jnsfZ: `Exception.class` also matches exception classes annotated `@ResponseStatus`, and `ExceptionHandlerExceptionResolver` runs before `ResponseStatusExceptionResolver`, so the annotation is never read:

```
@ResponseStatus(CONFLICT) exception, advices registered -> 500
same exception, advices absent                         -> 409
```

Consuming projects use `@ResponseStatus` on exception classes; two present examples are one annotated `@ResponseStatus(CONFLICT)` and one annotated `@ResponseStatus(BAD_REQUEST)`. This is the case the PR fixes for the Spring MVC exceptions — a declared status becoming 500 — applied to statuses declared by application code.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
jnsfZ: `fieldErrors` is gated on `include-message`. Spring Boot's property for this is `include-binding-errors`, which nothing reads — `src/testIntegration/resources/application.yaml:134` sets it and it has no effect.

So `include-binding-errors: never` still emits field names, and suppressing field names while keeping a readable `detail` is not expressible.

MIGRATION.md
jnsfZ: Bodies produced by the advices are `problem+json`. Responses written outside handler dispatch are not.

`JwtAuthenticationFailureHandler` calls `response.sendError(401, ...)`, which dispatches to `/error` and Boot's `BasicErrorController`, whose body is `{timestamp, status, error, path}`. Measured without a token on `/v1/roles`: 401, no `Content-Type`, empty body.

A missing or expired token is the most common client-facing error, and it carries no `type`, `instance` or `detail`.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The switch covers 404, 405, 406, 413 and 415. Everything else reaches `INVALID_INPUT`, including 401, 403 and 409 — while `AUTHENTICATION_FAILED`, `FORBIDDEN` and `DUPLICATE_RESOURCE` exist in `ErrorCode`.

`AuthenticationController` throws `ResponseStatusException` with `UNAUTHORIZED` (lines 140, 158, 179, 182) and `FORBIDDEN` (line 170). A failed login against the running application returns:

```
POST /auth/token, wrong password
401 {"detail":"Bad credentials","status":401,"title":"Unauthorized",
     "type":"urn:frachtwerk:error:INVALID_INPUT"}
```

Clients dispatch on `type`, so this does not match `AUTHENTICATION_FAILED`. A 403 raised as `NotAllowedException` and a 403 raised as `ResponseStatusException` also report different codes for the same condition.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ, line 302: Logging happens only for 5xx. The 4xx cases the parent routes here — media type, type mismatch, missing parameter/part/path variable, binding, `ResponseStatusException` — produce no log line.

The failed login above is one of them: a 401 from `/auth/token` is not logged at any level. With `include-message` at its default `never` the body is `detail: "An error occurred"`, so neither the response nor the log identifies what failed.

`AuthenticationController:138` also catches `AuthenticationException` broadly. `InternalAuthenticationServiceException` — raised when `loadUserByUsername` fails, e.g. an unreachable LDAP server or a database error — is a subtype, so an authentication backend outage takes the same silent 401 path.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ, line 382: The `ErrorResponse` branch requires a non-null detail. `ResponseStatusException` takes its detail from `reason`, so one constructed without a reason falls through to `exception.getMessage()`, which returns the status and the quoted reason:

```
ResponseStatusException(NOT_FOUND) -> detail "404 NOT_FOUND"
```

The framework's own exceptions supply a detail, so this concerns `ResponseStatusException` raised without a reason in application code — a common form. The method's javadoc notes that `getMessage()` carries the status prefix.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ, line 321: MIGRATION.md points applications at this method as the extension point. `handleExceptionInternal` and `validationResponse` call `problemDetailFactory.create(...)` directly, so an override applies to the handlers declared in this class but not to the types inherited from `ResponseEntityExceptionHandler` or to validation errors.

There is also no `HttpHeaders` parameter, so a 401 with `WWW-Authenticate` or a 405 with `Allow` cannot be produced through it.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ: The parent's `handleExceptionInternal` returns `null` when the response is already committed, resolves the detail through `MessageSource` when `body` is null, and delegates to `createResponseEntity`. This override does none of the three.

The first two are not reachable in the library as it stands — there are no async or streaming endpoints, and no `problemDetail.*` message keys are defined. The third applies now: `createResponseEntity` is named in the parent's javadoc as the hook for adjusting body, headers or status, and a subclass overriding it has no effect on any exception routed through here.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ, line 245: Neither branch logs the denial, and `ExceptionTranslationFilter` logs at TRACE, so a `@Secured` or `@PreAuthorize` rejection leaves no record.

The javadoc says this mirrors `ExceptionTranslationFilter`. The filter delegates to the configured `AccessDeniedHandler`, and to the `AuthenticationEntryPoint` for anonymous callers; neither runs here, so an application registering one does not see it used for denials raised inside the dispatch. No consumer configures one at present.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/FallbackExceptionHandler.java
jnsfZ, line 54: `Exception.class` also matches exception classes annotated `@ResponseStatus`, and `ExceptionHandlerExceptionResolver` runs before `ResponseStatusExceptionResolver`, so the annotation is never read:

```
@ResponseStatus(CONFLICT) exception, advices registered -> 500
same exception, advices absent                         -> 409
```

Consuming projects use `@ResponseStatus` on exception classes; two present examples are one annotated `@ResponseStatus(CONFLICT)` and one annotated `@ResponseStatus(BAD_REQUEST)`. This is the case the PR fixes for the Spring MVC exceptions — a declared status becoming 500 — applied to statuses declared by application code.

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/ProblemDetailFactory.java
jnsfZ: `fieldErrors` is gated on `include-message`. Spring Boot's property for this is `include-binding-errors`, which nothing reads — `src/testIntegration/resources/application.yaml:134` sets it and it has no effect.

So `include-binding-errors: never` still emits field names, and suppressing field names while keeping a readable `detail` is not expressible.

MIGRATION.md
jnsfZ, line 7: Bodies produced by the advices are `problem+json`. Responses written outside handler dispatch are not.

`JwtAuthenticationFailureHandler` calls `response.sendError(401, ...)`, which dispatches to `/error` and Boot's `BasicErrorController`, whose body is `{timestamp, status, error, path}`. Measured without a token on `/v1/roles`: 401, no `Content-Type`, empty body.

A missing or expired token is the most common client-facing error, and it carries no `type`, `instance` or `detail`.

jnsfZ (changes requested):

essencium-backend/src/main/java/de/frachtwerk/essencium/backend/controller/advice/GlobalExceptionHandler.java
jnsfZ, line 77: Is this also neccesary here?

pwannenmacher: Yes, it’s necessary. Without @Order, GlobalExceptionHandler would have a LOWEST_PRECEDENCE, just like the fallback advice. `OrderComparator` is stable; the relative order of two advices of equal precedence then depends on the bean detection order — which is undefined. If the fallback is sorted first, every exception becomes 500. This is not a theoretical risk, but the difference between ‘works’ and ‘happens to work today’.

jnsfZ (dismissed):

jnsfZ (dismissed):

jnsfZ (approved):

